PR #19, opened 2012-07-28
Got SparkleShare and git working somewhate

J5:


sonalkr132:
Hi! Git stuff is working now ( check #302) Sparkleshare is able to clone public projects as well. I am no expert on working on working of sparkleshare, my guess is that it is using polling for now. Any idea about how do I go about sending setting notification to notification server? @hbons  Would a symlink of this [post-hook](https://github.com/hbons/sparkleshare-git-hook/blob/master/post-update) be enough?

I have set up invite url as `sparkleshare://ruby-glitterfollowers.rhcloud.com/addie/cards/invite.xml`, however that doesn't work either. On chrome is launches `xdg-open` application. I am not sure what am I doing wrong here (besides not adding fingerprint)

Also, is there any way I can allow syncing of private projects as well and not have them ask my credentials every time sparkleshare tries to sync them.

hbons:
Here's a git-hook that posts to the notification service: https://github.com/hbons/SparkleShare/wiki/Notification-service#git-hook

sonalkr132:
Thanks @hbons It is the same hook I referenced.
is it possible to cache credentials for a sparkleshare client over smart http protocol?

hbons:
Oh, I missed the link in your comment. That's all you need to post a notification. Note the service is down at the moment. If it can't be reached SparkleShare falls back to polling. You can also specify your own notification service address in the invite XML.

If you're asked about credentials your SSH probably isn't added to your ssh-agent on login. You can also use the key provided by SparkleShare which doesn't use a password so won't ask you for one. SparkleShare doesn't support HTTPS.

sonalkr132:
> SparkleShare doesn't support HTTPS

It does. This might sound weird cause you wrote SparkleShare. It is able to fetch public repos over HTTPS. Only thing missing is that there is no way to cache HTTPS credentials. So every time SparkleShare tries to push to repo,  it promts for username and password. If I fill them correctly, SparkleShare is able to push as well. SparkleShare is using polling so I am asked to verify every 5 minutes or so and that is really annoying.

hbons:
There are many more issues with HTTPS in SparkleShare. It works sometimes, but it doesn't do anything like certificate validation, handling authentication, or errors. So it's not officially supported and is not recommended.

PR #181, opened 2014-12-14
Delete data

sbagroy986:
Get a Rugged::OS error (Failed to make directory 'public/data':File exists) if we don't deletw this file.


sarupbanskota:
Um, what exactly is happening here? I'm unable to follow..

EmilyDirsh:
For openshift, we should probably create the symlink in one of the hooks - only if it doesn't already exist. I'm going to close this for now.

PR #186, opened 2015-01-16
Don't allow following repeatedly

sarupbanskota:


rohitpaulk:
I haven't tried this out, but a quick look at the code and it looks good to me. Certainly better than what's in the current codebase anyway, so merging.

PR #192, opened 2015-01-24
Nuke Gemfile.lock, do a clean bundle

sarupbanskota:
**Please do not accept this PR until it's properly tested.**

So while at the design Hackfest, Emily and Zach had problems setting up GG locally, and it took us time to figure out what was wrong. I had set up GG on a variety of environments a week ago and had no problems, so it seemed strange to me. Now I'm beginning to find some insight.

The important problems: 

#### Inconsistent Ruby versions. 
* For me, I was on 1.9.3 all this time, because I was wrong in thinking hash syntax wouldn't be backwards compatible in Ruby 2 onward. On 1.9.3, our existing set up works completely fine.
* Now I have a fresh install, with 2.2.0, and I came across a host of problems. The old Gemfile.lock has references to "nio4r" (I honestly didn't know why), and nio4r [only supports until 2.1.0](https://github.com/celluloid/nio4r#supported-platforms). So, that explains the issue with Zach's setup.
* To experiment, I got on 2.2.0, nuked Gemfile.lock, removed version information from some Gems on the Gemfile, and rebundled from ground up. So naturally few Gems got upgraded to later versions, and the bundle went through comfortably without any errors at all. Thing to note: I wasn't building rugged off development branch, it was through whatever branch just `gem rugged` is pulling from. `gem list rugged` tells me `rugged (0.21.4)`.

#### Rugged oriented problem
* If I try to pull from development, it throws an error, which seems fair since I don't see any development branch on the rugged repo.
* If I'm on 1.9.3 and using the current setup, everything is good. 
* If I'm on 2.2.0 and adding a newfile to a new GG project, I get thrown an ``undefined method `lookup' for Rugged::Remote:Class`` error. I'm attaching a screenie as well.

Would some of you please try setting up GG on Ruby 2.2.0 and off this branch, and report if you have the same problem? If any of you have any insight on what could be wrong, please shoot, that saves me time I can use on fixing some features today :) Thanks!

screenshot from 2015-01-24 08 27 46


sarupbanskota:
@rohitpaulk A cursory lookup around rugged's history led me to this: https://github.com/libgit2/rugged/blob/8838ab5765cfd31763b784213bd64cc24d3ee84a/lib/rugged/remote.rb#L18. Could this be the problem (sorry I haven't gotten a chance to investigate in depth)?

rohitpaulk:
@sarupbanskota - Yep, looks like they've majorly simplified rugged's API. I'll have a look at this soon.

rohitpaulk:
I'd suggest not upgrading Rugged in this PR - have it stay at whatever the previous version was.

sarupbanskota:
Which is the version of rugged you use? (and the one that doesn't break on file upload?) I'll use that for now.

rohitpaulk:
My `Gemfile.lock` says `v0.19.0`. Although, the docs say that `v0.19.0` only supports Ruby 1.9. I used to have a Vagrant machine, but I don't remember what version of Ruby it was running. I guess there's no easy way out here :) 

Let's stick to the latest version, and I'll send in a PR to fix stuff up.

rohitpaulk:
I've restarted the build on Travis to make sure that our tests trigger this problem. (the initial build errored due to a timeout)

rohitpaulk:
@sarupbanskota - Can you add a `.travis.yml` and mention the ruby version to be 2+?

sarupbanskota:
@rohitpaulk I'll do it once I'm finished with some other branches I'm on.
btw, I pushed a related commit https://github.com/glittergallery/GlitterGallery/commit/6c74a024b0152067586d0d4ba2570d613bcb08d8 to your PR, so we can close this for now.

PR #195, opened 2015-01-25
Improve rugged 

sarupbanskota:
I'm attempting to clean up some of our old code.. I'm sending an early PR so the changes don't appear to be intimidating to review.
@rohitpaulk, here's some of the stuff I've changed (not everywhere, across files/methods I toyed with): 
* Removed all instances of Grit (most of them were in the repo-info/fork/pull-request oriented methods I had written, so I simply removed them for now).
* Tried to follow a `function_call arguments, more_arguments` style over one that uses braces, it feels cleaner
* Made hashes that went too long into multi-line, so they're easier to read
* Removed any irrelevant #FIXME/TODOs and added relevant ones
* Removed comments where the methods were self-explanatory.


rohitpaulk:
Other than the comment that I've left - everything LGTM.

rohitpaulk:
@sarupbanskota - A quick rebase on master, and the tests should pass :)

sarupbanskota:
@rohitpaulk actually, there's another problem, a hiccup with the urlbase method. If it's not a private repo, `self.uniqueurl.to_s` lends a `/` to the end, which doesn't match the expected path. I'll write a fix, few minutes.

sarupbanskota:
There you go. Fixed and green :) Let's not merge this right away, I intend on sending some more commits to this thread.

rohitpaulk:
Cool :)

sarupbanskota:
@rohitpaulk if you think it looks good, please merge it, I'll add more changes on another PR.

rohitpaulk:
I don't like the fact that we aren't storing `parent` as an integer. We should be storing `parent_id` as an integer and adding a `has_one` relation. Now that we've started using `parent` - I wouldn't change the db column, but we can still create the `has_one` relation and use it wherever possible.

sarupbanskota:
So, right now, here's what works (things could be inelegant, we'll clean up).
Do this:
* You want to test the tree view, so just take pick some git repo you have with directories and plug them into the data directory of a GG project you create.
* if you went to /sarup/funkyproject, you'll see stuff as usual, as I haven't taken that down yet.
* if you went to /sarup/funkyproject/tree/master, you'll see the directories and files listed. 
* /sarup/funkyproject/tree/master/level1/level2/leveln should work, they're also clickable on their pages
* /sarup/funkyproject/blob/master/path/to/file.rb should also work, displaying content of that that file.

Problems to fix:
* links with a / in the end have issues, when you click links via them, they break.
* the data of the blobs are just thrown plain, we want to represent in the svg form

I might be involved with another project for few days, in the meanwhile @rohitpaulk would you take up those two fixes and merge them in? 

also, in this flow, next few things we want: 
* ability to create directories from the web UI
* multiple file uploads

once we get that and ext public repo cloning, we're set from the git-side of things for our first test with the design team :) I'll open an issue to track the other stuff later.

sarupbanskota:
Update: 

Problems to fix:
* links with a / in the end have issues, when you click links via them, they break.
* <strike> the data of the blobs are just thrown plain, we want to represent in the svg form </strike>

rohitpaulk:
@sarupbanskota - Leaving this here in case you didn't get a chance to see it on IRC. 
```
11:08:56 PM	rohitpaulk  > display binary blob text as an image
11:09:02 PM	rohitpaulk  I thought we were doing that? in the history view
11:10:06 PM	rohitpaulk  https://github.com/glittergallery/GlitterGallery/blob/master/app/views/projects/file_history.html.haml#L19-L28
```

EmilyDirsh:
LGTM - merged

Issue #197, opened 2015-01-28
Trying to follow a user without logging in gives  401 unauthorized error


shubhpatel108:
If a user tries to follow another user without logging in, the server throws an 401 Unauthorized error. 
It should redirect user to login page instead. 
The root of the problem: The request is asynchronous and expect a JS response. We need to do more than just adding a authenticate_user filter (which renders HTML response of login page).

PR #198, opened 2015-01-29
Show followers

sarupbanskota:
@rohitpaulk +1?


app/views/users/_social.html.haml
```diff
@@ -13,17 +9,16 @@
         %p Follow
```
rohitpaulk: This isn't a problem that is new with this PR, I'm just leaving this here for reference. If we're going to show the "Follow" button to users who aren't signed in, we should show them a prompt to sign in if they click on it. As of now, it just results in a silent 401 (unauthorized) error.

shubhpatel108:
@rohitpaulk true. For the same reason, I created #197.
Let's show the follow option, and redirect the user to login page if he chose to do so.
Instead of changing the response type, we can even change the link to login page (on "follow").

rohitpaulk:
@sarupbanskota - This is definitely an improvement on what we currently have - but I see serious UX problems. 

1) When a user clicks on "Followers" or "Following", we hide the feed. The only way for the user to get this back, is to reload the page. I see that we have a tab that says 'activity', but that doesn't do anything yet. 

2) If I'm following a user and I click on "Followers", I see my name. When I click on unfollow, the count drops to zero but the list of followers is not updated.

I'm going to merge this in for now, we can always improve upon it later.

rohitpaulk:
@shubhpatel108 - Ah, sorry I missed that. Changing the link is a good idea, as long as we have a mechanism to ensure that the user ends up back on the original page after he logs in.

sonalkr132:
@rohitpaulk 
1. Which "feed" are we talking about? 
2. If I am on followers tab and I click on follow/unfollow(first option) should I move to follow/unfollow tab(again first tab)? I think better interface would be one where we click on follow/unfollow and it moves us to followers tab so that we can see new updated list, irrespective where we were previously.

Issue #209, opened 2015-03-03
Fix tests


rohitpaulk:
Current status on Master - 
`45 examples, 16 failures, 5 pending`

on #208 - 
`45 examples, 5 failures, 5 pending`

sarupbanskota:
@rohitpaulk good to close?

PR #210, opened 2015-03-03
Fix some tests

Alwahsh:
I've fixed some of the tests to fix issue #209 partially.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -64,8 +64,6 @@ class Project < ActiveRecord::Base
 
   # Push the existing contents of the satellite repo to the bare repo
   def pushtobare
-    barerepo = Rugged::Repository.new self.barerepopath
-    satelliterepo = Rugged::Repository.new self.satelliterepopath
     remote = satelliterepo.remotes['bare']
     unless remote
       remote = satelliterepo.remotes.create 'bare', barerepo.path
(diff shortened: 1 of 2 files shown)


app/models/project.rb
```diff
@@ -64,8 +64,6 @@ def satellitedir
 
   # Push the existing contents of the satellite repo to the bare repo
   def pushtobare
-    barerepo = Rugged::Repository.new self.barerepopath
-    satelliterepo = Rugged::Repository.new self.satelliterepopath
```
Alwahsh: I believe those 2 lines have no use, there are functions defined called barerepo and satelliterepo that return the same thing

app/models/project.rb
```diff
@@ -97,7 +95,7 @@ def init
         Rugged::Repository.init_at self.barerepopath, :bare
         Rugged::Repository.clone_at parent.satelliterepopath, self.satelliterepopath
       end
-      self.pushtobare
+      self.pushtobare unless satelliterepo.empty?
```
Alwahsh: This fixes 3 of the tests, we don't want to push something to a bare when it's already empty.

spec/spec_helper.rb
```diff
@@ -2,7 +2,6 @@
 ENV["RAILS_ENV"] ||= 'test'
 require File.expand_path("../../config/environment", __FILE__)
 require 'rspec/rails'
-require 'rspec/autorun'
```
Alwahsh: This is depreciated in the new version of rspec

rohitpaulk:
Great work! Travis isn't green yet, but I'm merging this - it's certainly better than what we currently have on master.

Alwahsh:
The other failing tests are probably related to issue #181. I'm trying to investigate about it

PR #211, opened 2015-03-03
Fix rest of tests

### LINKED ISSUE #209: Fix tests

Alwahsh:
This makes all tests pass... 
I think this line is useless... I can't see `@parent` being used anywhere in this file. If it's needed somewhere, I think we should move it to the controller. Project creation now succeeds.
Fixes #209 completely

### DIFF
--- a/app/views/projects/_project_toolbar.html.haml
+++ b/app/views/projects/_project_toolbar.html.haml
@@ -1,4 +1,3 @@
-- @parent = Project.find @project.parent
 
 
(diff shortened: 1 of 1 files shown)


sarupbanskota:
thanks! it looks like I (or somebody else) was using that for a temporary check.

PR #212, opened 2015-03-04
Fix issue of public projects not appearing

Alwahsh:
When viewing the '/inspiring' page, no one can see any public projects.
see 
SHA: a5c3785ed8d6a35868bc169f07e40e889087fd2e
 This is because we set value of private to true on creation when the user chooses to create a private project but do nothing to private when the user chooses public so its value remains nil. I've added a migration to make the 'private' column have a default value of false.
I've also added tests that would fail on previous commits but succeed with this one.

### DIFF
--- a/spec/models/project_spec.rb
+++ b/spec/models/project_spec.rb
@@ -22,6 +22,13 @@ describe Project do
   	expect(Rugged::Repository.new(File.join @project.path , 'satellite' , '.git')).to be_a(Rugged::Repository)
   end
   
+  it "gets list of inspiring projects" do
+	@project = FactoryGirl.create(:project)
+	FactoryGirl.create(:project, :name => 't2', private: true, user: @project.user)
+	expect(Project.inspiring_projects_for(@project.user.id).count).to eq(0)
+	expect(Project.inspiring_projects_for(@project.user.id+1).count).to eq(1)
+  end
+
   describe ".urlbase" do
   	it "is correct" do
   		@user = FactoryGirl.create(:user, :username => "sarup")
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -19,7 +19,7 @@ class ProjectsController < ApplicationController
 
   #TODO Limit to popular/recent projects
   def index
-    @projects = Project.where.not(private: true, user_id: current_user.id)
+    @projects = Project.inspiring_projects_for current_user.id
   end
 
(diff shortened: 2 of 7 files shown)


spec/models/project_spec.rb
```diff
@@ -22,6 +22,11 @@
   	expect(Rugged::Repository.new(File.join @project.path , 'satellite' , '.git')).to be_a(Rugged::Repository)
   end
   
+  it "lists public projects for a certain user" do
```
rohitpaulk: How exactly does this test case test "lists public projects for a certain user"? 

It tests the query `Project.where.not(private: true)`. If we wanted to test that, (have this as a unit test for the integration test above), then this query should be moved in to the Project model, so that we're testing the same code that the controller uses :)

rohitpaulk:
LGTM, except for that one comment.

spec/models/project_spec.rb
Alwahsh: Yeah, I'm sorry... The test was different at first then I changed it and forgot to change the name :). I think I should just change the name as the purpose of this test was just to make sure public filtering works.

Alwahsh:
Do you prefer if I squash the commits?

rohitpaulk:
@Alwahsh - I guess I wasn't clear before, my problem isn't with the **name** of the test, it's with the **purpose**. This test isn't really testing 'application' code, it's testing a query that is defined within the test. What happens if I send in a PR that changes 'projects#index' to use a different query? The test would still pass. 

Tests should test application code, not code defined within them. Two solutions I see here - 

- Break out the query into a method on the Project model (such as `Project.projects_for_index`), and use that both in the projects controller **and** the test. 
- Move this test to a controller test, and check whether the `@projects` variable is assigned properly in `projects#index`. 

However it's done, we should only have one query - and it should be used by both the test and the application code. 

@Alwahsh - I really appreciate the work you're doing and I'm sorry to block your PR based on the structure of tests, but I hope you find my reasons reasonable :)

Overwriting/squashing your commits would be good, yes.

Alwahsh:
@rohitpaulk  I think I got your point :).
Yes, I totally agree... I just used to write a test that fails at this bug to help me solve it and prevent its occurance in the future but I can see now that it's better to test the real code in the app that caused the bug. Thanks :)

rohitpaulk:
That's perfect. Will merge once Travis signals green.

PR #213, opened 2015-03-04
Fixed 'No-Image' Uploading

sbagroy986:
If one clicks on the 'Upload File' button without actually selecting an image to upload, we get this:  http://gyazo.com/1f5511b5b1ef5cda7f51c158939e9f39

After this, we get:  http://gyazo.com/465bb476c1e58852eeda2a6bb02536f9
We could change the message to 'No image selected', but i'm not sure what other cases there can be when 'params[:file]' is nil.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -247,10 +247,10 @@ class ProjectsController < ApplicationController
 
   def file_upload
     @project = Project.find params[:id]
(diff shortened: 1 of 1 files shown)


rohitpaulk:
> We could change the message to 'No image selected'

Yes, do change the message. There aren't any other cases that I can't think of.

Issue #200, opened 2015-02-02
Dealing with thumbnails


sarupbanskota:
What are your opinions on this? Do we generate thumbnails per commit? Do we save them within the satellite repo? Do we save them elsewhere? I wouldn't keep them on the satellite repo, and would create a thumbnails dir along with satellite and bare repo. For each commit that takes place, I'd generate thumbnails and place them within a dir with the same tree hash as that of the commit's. 

The other option is to do thumbnails on the fly.. ie on the thumbnails dir, generate thumbnails as the page request is made, but that would be a slow approach imo. I've writing this super simple gem called `phototrim`, we could use that for now.

EmilyDirsh:
IMO, thumbnails are for GG use, and we don't need to pollute the repo with them. +1 to the separate thumbnails dir and generating and saving a thumb whenever a commit is made. We can use the commit id to name them.

EmilyDirsh:
I believe this fixed with the referenced PRs.

PR #214, opened 2015-03-04
Add thumbnails generation on new commits

Alwahsh:
A new directory is created on project creation for holding the thumbnails.
Thumbnails are generated on each commit and added to this directory under the same commit id.
The size of the thumbnail is set to 50x50 for now but I think it's better to allow that to be modified in a configuration file or admin panel.
Thumbnails are only generated for now. They're not shown yet.
targets issue #200.


app/controllers/application_controller.rb
```diff
@@ -15,12 +15,18 @@ def return_current_user_projects
     end
   end
 
+  def generate_thumbnail(path, imagefile, commit_id)
+    image = Magick::Image.read("#{path}/satellite/#{imagefile.original_filename}").first
+    image.scale(50,50).write "#{path}/thumbnails/#{commit_id}"
+  end
+
```
Alwahsh: I think these methods should be present in projects_helper instead of being in application_controller. They are only used in the projects_controller as far as I can tell.

EmilyDirsh:
LGTM - merging.

app/controllers/application_controller.rb
sarupbanskota: yup, you're right - you can move them to the projects controller. i'd love it if you could also watch out for similar nits and move them to the better location at your discretion. just remember to leave a note on those when you send PRs, so we don't overlook - for most of the part though, travis should tell us :)

app/controllers/application_controller.rb
```diff
@@ -15,12 +15,18 @@ def return_current_user_projects
     end
   end
 
+  def generate_thumbnail(path, imagefile, commit_id)
+    image = Magick::Image.read("#{path}/satellite/#{imagefile.original_filename}").first
+    image.scale(50,50).write "#{path}/thumbnails/#{commit_id}"
```
sarupbanskota: the thumbnail size is something we might want to revisit - we'll have variable sizes of images and setting a fixed dimension might not be the best approach. ofc, it's better than what we have atm, just noting.

Alwahsh: Yeah, I'm planning to do that as I said in the description :)

app/controllers/application_controller.rb
Alwahsh: Alright. Thanks :)

PR #215, opened 2015-03-04
Multiple-file uploads

sbagroy986:
Added multiple-file uploading by Ctrl+selecting images.


EmilyDirsh:
Could you add a test for this? spec/features/projects_spec.rb has a test that involves uploading a single file - you should add a similar one that uploads multiple files.

sbagroy986:
I don't really have much experience with rspec, but there are already 42 failures when I run it. I'm not sure we should add another test before sorting those out.

Alwahsh:
@sbagroy986 did you run 
RAILS_ENV=test rake db:schema:load ?

sbagroy986:
I did, right after I replied. But there's still this:  http://gyazo.com/c09707f7d8d36ad313ecb6d6a168bd6d
I'll try fix it.

Alwahsh:
@sbagroy986 I think you've got a version that didn't include this file... Simply copy the file happypanda.png from the public folder into the path specified in the error :)

sbagroy986:
I already have the file in the path specified. So that doesn't seem to be the case.

Alwahsh:
Right... Sorry. I misread the error :)
open:
'spec/features/projects_spec.rb' at line 41
replace it with this:
`page.attach_file("file[]", 'spec/factories/files/happypanda.png')`
this is because you changed the name of the field from "file" to "file[]".

sbagroy986:
Yeah I realized that soon after. Thanks though :)
Are there any particular images I should use for the tests? If not, the tests are done.

app/views/projects/newfile.html.haml
```diff
@@ -30,7 +30,7 @@
       %div
         - if current_user.id == @project.user.id
           = form_tag file_upload_project_path(@project), multipart: true do
-            = file_field_tag "file" 
+            = file_field_tag "file[]", multiple: true 
             = submit_tag
```
rohitpaulk: We shouldn't have to make this change in two places - would be great to extract this into a common file. Just leaving this as a note, PR looks good to merge.

Alwahsh:
@sbagroy986 I understand the reason for the naruto.png in the tests folder but why is there one in the public folder? I can't see any use for it.

sbagroy986:
happypanda.png is also present in both the places. I assumed there was some use for it there and put naruto.png in there as well.

@EmilyDirsh Are there some other changes to be made or is this ready to merge?

Alwahsh:
@sbagroy986 happypanda.png was already existing in the public folder to be used as the default avatar if none is found for the user in gravatar. I needed some image for testing so I copied it to the tests folder.

sbagroy986:
File removed from the public folder.

sarupbanskota:
Merged - thanks!

PR #218, opened 2015-03-05
Improve thumbnails generation and show thumbnails in logs

Alwahsh:
These are some improvements to the thumbnails generation.
The size of thumbnails is now configurable from the '0_glitterconfig.rb' file.
There was a problem that when updating files, no thumbnails are generated. Fixed that.
Thumbnails are now shown in the logs page.
I think this resolves issue #200 completely


sarupbanskota:
@Alwahsh, could you pull from your end once, and resend a conflict fixed version?

Alwahsh:
@sarupbanskota  done :)

PR #223, opened 2015-03-07
Added file validation

sbagroy986:
Checks format by getting the magic number of the file.

PS: Slight error in one of the commit messages. This is not client side validation.


Alwahsh:
'ruby-filemagic' needs libmagic-dev.
you need to add these lines in .travis.yml to make it able to install the gem.
```
before_install:
  - sudo apt-get update -qq
  - sudo apt-get install -y libmagic-dev
```
I also suggest that you squash your commits to get rid of unnecessary history :)

sbagroy986:
Didn't check back to see the Travis build had failed. Thanks @Alwahsh :)
I'll squash them in a bit. In the middle of something right now.

PR #225, opened 2015-03-07
removed fb app key and secret

sonalkr132:


rohitpaulk:
That was quick :wink:

PR #226, opened 2015-03-07
Upgrade to rails 4.1 and a few other bug fixes

sonalkr132:
secret.yml has better way of handling with app secret keys than earlier version of rails. My main aim was that we don't need to save production secret key in repo. 
I have used the same keys as defaults so that current session running locally in dev env don't end up being invalid after a fetch.  

Suggestion requests:
* Do we need to make changes to installation readme? Ideally we should ask them to generate new secret using "rake secret" but that's not entirely necessary.
* Rails 4.1 doesn't support mulit json and things like return statements in code blocks. I just want to make sure that I am not breaking anything [further reading] (http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-0-to-rails-4-1)

UPDATE 1 :
* changing "http://rubygems.org" to "https://rubygems.org"  fixed "Bundler::MarshalError TypeError: incompatible marshal file format" error during bundle install
* As I said upgrading to rails 4.1 may break something.. they removed defualt_scope without block, which was picked by travis. I have fixed that, but there might be something in untested part of code.

UPDATE 2:
* New comments were giving "NoMethodError - undefined method `urlbase' for nil:NilClass" because of line:
```Ruby
= link_to "Issue?", "#{@project.urlbase}/issues/new?desc=#{comment.body}"
```
Project has to be an instance variable if view has to pick it up.


EmilyDirsh:
I'm going to close this, since these commits are also included in #231

Issue #230, opened 2015-03-08
irreversible migrations


sonalkr132:
"20150304042821_make_projects_public_by_default" was irreversible, I fixed that. "20140518213805_add_devise_to_users" is also irreversible. I am not sure if I should make changes to its down method. Are we using it as fail safe?

EmilyDirsh:
20140518213805_add_devise_to_users is nearly 10 months old. I wouldn't worry about making that reversible.

Issue #229, opened 2015-03-08
Implementation of project.parent


sonalkr132:
It started with: "ActionView::CompiledTemplates::FORK" error on issue/show page
```Ruby
- if @project.id == @project.parent
          %span.badge.author PARENT
        - else FORK
```
As you can tell, it's and easy fix:
```Ruby
       - else
         %span.badge.author FORK
```
But There is lot more to it than that. Project create has used:
```Ruby
def create
    project = Project.new project_params
    project.user_id = current_user.id
    project.parent = project.id
```
project.id doesn't get created until after object is saved in database, so all the project.parents are just nil. I need suggestion solving this.
Now we can write that line
```Ruby
    project.parent = project.id
```
after 
```Ruby
    project.save
```
I don't that that would be very good idea. We would already be going against rails convention and in fork method we are setting child.parent = Project.parent. So every fork will end up having same parent.

Is thats what you guys want? Doesn't git work like  if I fork from a original_repo and if someone forks my repo then their parent should be my repo not the original repo. Confusing?
What I am suggesting is that parent relationship should be like:
original_repo <- my_repo <- my_repo_fork

what we can do is in create set project.id = nil
and in fork we use the id of project from which we are forking and not the parent.id of project being forked.

Let me know what you guys had on your mind and I will send you a PR.

rohitpaulk:
Aah, thanks for the analysis. I agree with you, let's set project.parent to `nil` for a non-fork, and to the id of the project being forked (not the parent, the id of the object itself) for forks.

sonalkr132:
Okay. should I migrate all the way down and make changes to "20131030132834_add_parent_to_projects" migration or should I add a new migration? I think you meant former?

rohitpaulk:
Ah I'm sorry, I didn't think the change through - the migration should be a simple rename of a column, and hence won't change existing data. What I meant was - when you add the new migration, it's okay if it doesn't honor the values on previous projects. 

TLDR - add a new migration that renames the column.

EmilyDirsh:
Actually, if you're forking a project from a fork, the parent of the new project _should_ be the parent id not the project id. That's not very clear. :P Take this diagram:
```
 ___________________________________________________
|   Project A [not a fork] id: 1, parent id: nil    |
[...]

PR #231, opened 2015-03-08
relation between forks and repos defined

sonalkr132:
As suggested, I have changed parent to parent_id and defined relation between forks and it's parents. `project.forks` and `project.parents` commands work from rails console. They would be handy, when we implement view for seeing all the forks of a certain project.
I had to make a few changes in view as per new project<-fork relationship.

I tend to follow thumb rule that whenever you add a migrate you try to migrate it all the way down to version 0. I came across many irreversible migrations, I have fixed some of them but eventually I had to give up and comment things out whenever reverting failed.. I would like to discuss how we should handle this irreversible migration situation here #230


app/controllers/projects_controller.rb
```diff
@@ -305,7 +304,7 @@ def fork
     child = Project.new name: @project.name,
                         user_id: current_user.id,
                         uniqueurl: @project.uniqueurl,
-                        parent: @project.parent
+                        parent_id: @project.id
```
EmilyDirsh: The problem with this, and I mentioned it in my comment to #229, is that if you're forking a project that is _itself_ a fork, the parent of the new project should be the original project, not the fork. Setting a non-forked project as its own parent, and using the parent id as the parent of forks accomplished this.

app/models/comment.rb
```diff
@@ -3,7 +3,7 @@ class Comment < ActiveRecord::Base
   belongs_to :user
   validates :body, :presence => :true
 
-  default_scope order: 'comments.created_at DESC'
+  default_scope { order('comments.created_at DESC')}
```
EmilyDirsh: Why did you change the coding style here? It doesn't match the coding style in the rest of the project.

sonalkr132: Support for default_scope without block was removed from rails 4.1. You should look [here](http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#changes-on-default-scopes#changes-on-default-scopes) and [here](http://stackoverflow.com/questions/24163571/default-scope-activerecord-4-1-block-usage-for-ordering#answer-24164710) 
Although, doc of default_scope still says we can use ones without block. But error in log clearly said that default_scope without blocks is no longer supported. I think that stackoverflow question explains it a bit.

EmilyDirsh:
@sonalkr132 can you fix the merge conflicts and update the PR? Thanks

sonalkr132:
Why did we need to remove:
```Ruby
issue = Issue.find(@comment.polycomment_id)
project = issue.project
```

Alwahsh:
@sonalkr132 
I simply removed 
```
issue = Issue.find(@comment.polycomment_id)
project = issue.project
```
because I inserted another hiddent field that held the project... No need to search for it this way anymore...

sonalkr132:
It doesn't solve the problem of ajex call not succeeding. view uses this line
```Ruby
%footer
    %p
      = link_to "Issue?", "#{@project.urlbase}/issues/new?desc=#{comment.body}"
```
Now unless you define that project variable in create method, call won't succeed.

PR #235, opened 2015-03-09
Fixed Error if No Project Exists (URL)

mecyborg:
If someone access the URL as /username/any_project and that project doesn't exists, the application was reporting error.

Update/Fix: Added a check and redirect to dashboard if project doesn't exist.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -129,10 +129,6 @@ class ProjectsController < ApplicationController
 
   def show
     @images = []
-    if @project.blank?
-    redirect_to dashboard_path
-    flash[:alert] = 'The project does not exist.'
-    else
     barerepo = @project.barerepo
     unless barerepo.empty?
       headtree = barerepo.lookup barerepo.last_commit.tree_id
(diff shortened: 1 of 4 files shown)


app/controllers/projects_controller.rb
```diff
@@ -129,6 +129,10 @@ def show_tree_content
 
   def show
     @images = []
+    if @project.blank?
+    redirect_to dashboard_path
```
rohitpaulk: Code indentation isn't proper, and you haven't closed the `if` condition with an `end`. Moreover, this isn't the right way to approach the problem. If the project doesn't exist, we should throw the error right in `return_context` and render a 404.

You've also checked in the `hot_deploy` marker file.

app/controllers/projects_controller.rb
```diff
@@ -129,6 +129,10 @@ def show_tree_content
...
+    flash[:alert] = 'The project does not exist.'
+    else
```
EmilyDirsh: Small nitpick - fix the indentation of the nested statements, so it's a little more readable.

app/controllers/projects_controller.rb
mecyborg: I was also wondering whether to render 404 or redirect to dashboard. But, now it clear, thanks :)
I'll fix the code indentation and if statement.
hot_deploy shouldn't be there. No worries, I'll fix that too if it there.

rohitpaulk:
If you decide to correct this PR and push again - please add a failing test.

mecyborg:
This works fine according to me. Please verify :)

rohitpaulk:
@mecyborg - Please add a failing test so that this error doesn't pop up again in the future.

mecyborg:
@rohitpaulk Failing test added.

PR #236, opened 2015-03-09
Fixed update option

### LINKED ISSUE #234: Users can update pictures in projects of the other users
Basically, I can edit any project of any other user right now because it gives me the 'Update' option.
eg:
8935c981c685653400236da20e72a937

I am logged in as 'sba23' but can update a picture in the user 'sbaa23's repo.
Is this how its supposed to work? I already fixed this locally but I wasn't too sure how the system is supposed to work.

sbagroy986:
Fixes #234

### DIFF
--- a/app/views/projects/masterbranch.html.haml
+++ b/app/views/projects/masterbranch.html.haml
@@ -29,8 +29,9 @@
       %div.photo
         %p{title: params[:image_name]}
           = image_tag @imageurl.gsub("public","")
-        %form{ action: File.join(params[:image_name], "update")}
-          %input{ type: "submit", value: "Update"}
+        - if @project.user_id == @current_user.id
+          %form{ action: File.join(params[:image_name], "update")}
+            %input{ type: "submit", value: "Update"}
 
     %section.comments
       = render partial: "comments/comment", collection: @comments
--- a/app/views/projects/update.html.haml
+++ b/app/views/projects/update.html.haml
@@ -24,11 +24,12 @@
     = render 'project_toolbar'
     = render 'shared/messages'
 
-    %section
-      %div
-        = form_tag file_update_project_path(@project), multipart: true do
-          = file_field_tag "file" 
-          = text_field_tag "message","commit message"
-          = hidden_field_tag "image_name", params[:image_name]
-          = submit_tag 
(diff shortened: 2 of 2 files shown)


Alwahsh:
while this might work in the short term, I think eventually we'll need to use an authorization gem... cancancan for example.

mecyborg:
I guess this is just a temporary solution to this. There should be a solid authorization for these type of bugs. What do you think ?

rohitpaulk:
Woah, that is not good. 

We have a check in https://github.com/glittergallery/GlitterGallery/blob/master/app/controllers/projects_controller.rb#L27-L34, could you extract that out into a method like `return_context` and make sure it's called before actions like `delete`, `file_upload` etc?

sbagroy986:
@rohitpaulk  okay sure.

@Alwahsh  I wouldn't suggest cancancan. Its a pain. Speaking from experience.

Alwahsh:
I've used it in multiple projects before and it was reliable...

sarupbanskota:
Yup, extract those relevant bits out - we can remember to keep a check on it's inclusion the next time a PR is sent around the sensitive methods.

EmilyDirsh:
Yeah, this should be in a before_filter or something similar. I like the idea of using an authorization gem for this - as the app grows, we're going to need some RBAC (role-based access control ;)) @sbagroy986 if cancan is problematic, do you have another suggestion?

sbagroy986:
Actually,CanCan doesn't support rails 4.
CanCanCan, does. I've had some pretty bad experiences with it though. 
There a few alternatives but I haven't personally used any of them so I don't have a suggestion. In case I'm the only one who has hade problems with CanCanCan, then I guess we can go ahead with it.

sbagroy986:
I saw a CanCanCan PR around. Do we close this then?

sarupbanskota:
Could you please link to a demo? I'm leaning towards merging this if it looks good since it's better than what we already have, and then switch to cancan when it's ready.

rohitpaulk:
This is an important change, and should not be merged without concrete tests to prevent it from happening again.

sarupbanskota:
@sbagroy986 any updates with the tests?

PR #240, opened 2015-03-10
Add fork network.

### LINKED ISSUE #237: [RFE] Fork "network" view
Given the changes in #231, it'd be nice to have a way to display the relationships between an original project and all the downstream forks of that project, whether forked from the original or another fork. 

This is similar to the GitHub network feature.

I'd recommend the [Ancestry](https://github.com/stefankroes/ancestry) gem for the implementation.

Alwahsh:
Use Ancestry gem to introduce a Fork network view.
Also, added the parts needed to activate delayed job to send mails in the background.
Resolves #237 , #252


EmilyDirsh:
@Alwahsh, do you have this running on your demo on OpenShift?

Alwahsh:
@EmilyDirsh I didn't but now I do :)
sign in with this info:
name@name.com
12345678
http://glittergallery-themonster.rhcloud.com/forker1/The%20Root/network
also try to view one of the projects as it shows now its parent with link to it.

Alwahsh:
I've done the change as you requested. Please let me know if there's something else :)

EmilyDirsh:
This is great! Thanks for getting it up on your demo site. 

@sarupbanskota, @rohitpaulk can you also review this and check out the demo and see if you have any feedback. Thanks!

The only thing I noticed that I think we should handle is someone trying to fork a repo they've already forked. For instance, I fork forker2/root_project, then try and fork root/root_project. It fails, becuase myusername/root_project already exists, but it just presents a generic error message. The same thing would happen if you tried to fork a project with the same name as a project you already own. I'm actually not sure if this is something we should try to enable, or just create a better error message for. At least short term, a better error message would be an improvement.

Alwahsh:
@EmilyDirsh one way to enable that is to make the urls of projects something like:
forker2/{project_id}-root_project
so instead of just making the url contain the project name, we make it contain the id too to make it unique. Not sure though that this would be desirable.

EmilyDirsh:
@Alwahsh that would be one way to do it, but like I said, I'm not really sure we should. I don't actually see much need for it, personally. I _do_ think we need a better error message, though.

app/controllers/projects_controller.rb
```diff
@@ -318,8 +322,8 @@ def fork
       redirect_to child.urlbase
       # todo - notifications
     else
-      flash[:alert] = "Couldn't fork project, try again."
-      redirect_to @project
+      flash[:alert] = "Couldn't fork project, you already have a project with the same name."
```
Alwahsh: @EmilyDirsh do you think it's safe to assume that the cause of the error here would be that he has a project with the same name... Since the parent project is valid, I'd say that this is the only cause for an invalid project.

sonalkr132:
You can view of projects a little DRY. 
```Ruby
%header
       %h1.generated
         = avatar @project.user.email
         %p
           = link_to @user.username, "/#{@user.username}"
           = "/"
           %span 
             = link_to @project.name, "/#{@user.username}/#{@project.name}"
       %h2 
         - if @project.root?
           %span.badge.author
             PARENT
         - else
          %span.badge.author FORK
```

Move it to a partial like "_project_header" and you will take away 50-60 lines of code. 
I have left out `%article` cause it has some issue. look [here](http://stackoverflow.com/questions/28939336/nested-render-in-haml)

I had done this in my PR [ab51dbd1053](https://github.com/sonalkr132/GlitterGallery/commit/ab51dbd105335c0c2cf2f7fce584961ecb13eb62), but since you are working on it, I have closed it so that we don't have merge conflicts.

app/controllers/projects_controller.rb
rohitpaulk: It's possible to look at child.errors and figure out, right?

Alwahsh: @rohitpaulk yeah it's possible but do we need to? ;)
I can't think of real case where the validations could fail except that.

rohitpaulk: @Alwahsh - fair enough, makes sense.

EmilyDirsh: @Alwahsh it could happen because of a server failure - i.e. couldn't connect to the database; undiagnosed server bug; etc

EmilyDirsh:
Good point @sonalkr132 - also we most definitely should DRY up badge handling. Twelve of the altered files in this PR are just updating badge handling, according to my count. That's kind of ridiculous. That should probably be a separate issue/PR though.

Alwahsh:
@sonalkr132 I agree with Emily. I had seen that change in your PR. I prefer that you do it as a separate PR later :)

app/controllers/projects_controller.rb
Alwahsh: @EmilyDirsh at that point, you'd see an error page not the normal page with a flash message :) Anyways, I'm planning to do it in a better way :)

Alwahsh:
@EmilyDirsh I've updated the PR with your suggestion including your comment on #237 . 
Projects are now soft deleted.
Please use the same info to access the page and see how the network appears... I've deleted project [The Root](http://glittergallery-themonster.rhcloud.com/aa/The%20Root). You might want to visit that to see how the page responds to deleted projects ;).

Please let me know if there's anything else you want to be done :)

sonalkr132:
Hey, shouldn't we update wiki that we have delayed jobs?

EmilyDirsh:
@sonalkr132 yep, there's a lot of stuff we need to update on the wiki :P

PR #242, opened 2015-03-10
Update demo link in Readme

Alwahsh:
Not much of a change but since we have a new demo now, let's put it in the readme :)


sonalkr132:
nice! :thumbsup:

PR #244, opened 2015-03-10
Bug fixes on Image upload and CanCan fixed.

charithhewage:
Bug fixes on Image upload and code refactored.


Alwahsh:
@charithhewage  you're using cancan which is not compatible with Rails 4.
Consider switching to [cancancan](https://github.com/CanCanCommunity/cancancan) :)

charithhewage:
@Alwahsh Thank you for your comment and also the good news. 
I changed it to cancancan.
As mentioned in the wiki most of the functionalists are same as cancan. Isn't it? :)

Alwahsh:
yeah but cancan is no longer maintained :)

charithhewage:
@Alwahsh: Can you kindly review my code and give me a feedback?

Alwahsh:
@charithhewage 
Just to make it clear... I'm not an official maintainer (yet :P).
Let me have a quick look and point you to errors I see.

Gemfile
```diff
@@ -30,7 +30,7 @@ gem "oa-openid", :require => "omniauth/openid"
 gem 'activerecord-session_store'
 gem 'actionpack-action_caching'
 gem 'rails-observers'
-
+gem "cancancan"
```
Alwahsh: I'd rather specify versions of gems when I use them. However, this is not applied in the rest of the file so I think it's OK.

app/controllers/projects_controller.rb
```diff
@@ -206,6 +206,8 @@ def projectcommit
   #todo - /tree/master/file_name and generalize for /tree/branch/file_name
 
   def masterbranch
+    authorize! :update_image, @project
```
Alwahsh: I don't see a reason for this being here... This prevents everyone from entering into the image page(I actually tried it locally and it really blocked other users).. I'd just prevent them from seeing the update button.
So, remove this line.

Alwahsh:
OK... so along with my comments inline... I see you've hidden the upload section but if someone managed to send a request to the app, he'd still be able to upload so I suggest authorizing the `file_upload` method in the projects controller as well.
Also, adding tests that verify your changes is always a good idea :)

charithhewage:
Ok thank you for your kind attention. Most of the codes should be refactored. I am still working on this.
Also the test cases should be written :)

app/controllers/projects_controller.rb
charithhewage: With reference to this issue https://github.com/glittergallery/GlitterGallery/issues/234

Alwahsh: That actually eliminates the access to the page itself not just the update button :)

Gemfile
charithhewage: @Alwahsh : You are correct. As a best practice we should specify gem versions on gemfile to avoid unexpected breakdowns while updating the bundler. I am working on this. thanks for the comment :)

sarupbanskota:
Just updating - the status of #236 depends on this PR.

charithhewage:
@sarupbanskota 
With this cancancan integration, most of the places were refactored. I am working on this.

charithhewage:
@sarupbanskota 
Sorry for the delay. Suddenly I got sick and little late to reply.  
I have done some changes of this PR and written test cases as well.
Also i am trying to re-factor routes page in routing branch (WIP).

Alwahsh:
@charithhewage you might want to check #233... We don't want to be working on the same thing ;)

public/403.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>The change you wanted was rejected (422)</title>
```
Alwahsh: I wonder why you put a 422 page to respond to unauthorized access.

charithhewage: cancancan redirects to 403 page by default if it's an unauthorized access.

Alwahsh: cancancan actually raises an exception in that case by default but according to the lines you added in the application_controller:
```ruby
rescue_from CanCan::AccessDenied do |exception|
  render :file => "#{Rails.root}/public/403.html", :status => 403, :layout => false
end
```
You'd get the 403 redirection effect but I'm talking about the content of the 403 page... The content here seems to be a response of 422... At least that's what the user would understand if he arrived here.

charithhewage: Sorry @Alwahsh,
That was my mistake. I copied the content from 422.html. Updated the code. Thank you for your review.

charithhewage: I am really happy to work with you @Alwahsh  :)

Alwahsh: Thanks @charithhewage :) (Sorry for my late reply, just saw it :))

sarupbanskota:
I'm wondering if this is complete and is pending review? 

https://github.com/glittergallery/GlitterGallery/pull/292 is waiting on this.

sonalkr132:
I tried to solve merge conflicts but after half an hour I was able to rebase only 5 commits, this PR has 18 (way too many). It is also outdated. Most of the changed files don't exist anymore.

 I am going to start on my own. However I am confused between [pundit](https://github.com/elabs/pundit) and [cancancan](https://github.com/CanCanCommunity/cancancan)

sarupbanskota:
@Aditya, I'd recommend CanCanCan, since iirc you previously mentioned you
have experience working with it + it's the better known one, so it's
comfortable to continue working with for potential new folks.


Sarup

PR #250, opened 2015-03-11
Improved grouping of the gems, omniauth only for production env and improved design for notifications

philippneugebauer:


EmilyDirsh:
Hi Phiipp, Omniauth is used in our test suite, so it needs to be available in our test environment - not just prod. That's why the tests are failing for this PR.

philippneugebauer:
I will start again from scratch, that's easier

Issue #185, opened 2014-12-30
Simplify comments using polymorphic associations syntax


rohitpaulk:
We have the schema for a polymorphic association (`polycomment_id` and `polycomment_type`), but we aren't making use of the rails syntactic sugar when working with comments in controllers.

http://guides.rubyonrails.org/association_basics.html#polymorphic-associations

sarupbanskota:
@rohitpaulk I'm thinking we can assign this bug to the next person who asks for a starting point.

rohitpaulk:
I don't think that we should 'assign' simple bugs at all. If we start assigning people bugs this small, it's just going to be more work for us - pinging them to ask if they've worked on it etc. 

Assigning does make sense for bigger features, stuff that would take time to work on :)

sarupbanskota:
by *assign*, I meant more like direct them to this, we don't follow up with them if there's no news of course.

or are you suggesting we look after these bugs ourselves?

rohitpaulk:
> by assign, I meant more like direct them to this, we don't follow up with them if there's no news of course.

Oh, then it's ok. I thought you were talking about locking issues and preventing others from working on them. 

> or are you suggesting we look after these bugs ourselves?

No no, not at all :)

rohitpaulk:
https://github.com/glittergallery/GlitterGallery/pull/254#issuecomment-82853815

Closing.

Issue #232, opened 2015-03-09
New Comment AJAX not working


mecyborg:
After adding new comment, it is not updated. Have to refresh the page to view new comment.

sbagroy986:
Okay so the problem here isn't the AJAX. Its most probably the ActionMailer. 
If you comment out the following chunk of code
![screenshot from 2015-03-09 15 21 36](https://cloud.githubusercontent.com/assets/7946556/6552814/0d21dd40-c670-11e4-9c0f-2fa6db8f1f9a.png)

and the "Issue?" line in _comment.html.haml,  the comments work fine. 
I'll try and look into it in a bit.

mecyborg:
Comments are working fine. Its just the js which is not rendering.

sbagroy986:
I meant the comments would render if you do what I mentioned above :)

mecyborg:
Ok, thanks ! I guess this is the ActionMailer as you said. :+1:

sonalkr132:
I have already fixed it, look at #226 
Now, there is another thing which is not working.. that is comments from others on your on your repo. This one might have something to do with actionmailer. I guess we can look into that?

EmilyDirsh:
The notification should be triggered in the background, and non-blocking. Anyone who wants to tackle this, just claim it.

Alwahsh:
@EmilyDirsh Do you mean creating the notification object itself should be done in the background or sending it to the user? if you mean the latter, then it's already done in #233 :)

sonalkr132:
I have suggestion related to this. Comment made on someone else's repo don't show right away, cause development.rb isn't configured properly for sending mails. [letter_opener](https://github.com/ryanb/letter_opener) handles it quite well. It sends kinda stubed email right in the browser and doesn't actually send it. I integrated it this app and it solved the issue of ajax call not completing. 
What do you guys think?

EmilyDirsh:
@Alwahsh I meant sending the email to the user. The non-blocking part is important too. Some environments may not have working outgoing email, so we don't want to block the rest of the execution because email doesn't work.

Alwahsh:
@EmilyDirsh Great.. That's exactly what I did using Delayed job. The PR is not ready to be merged yet, but I'm trying to finish it ASAP :)

EmilyDirsh:
[...]

PR #254, opened 2015-03-12
Fix ajax comments

sonalkr132:
* Had to change "project" in comment controller to instance variable for view to pick it up.
* added mailcatcher so that no config is required for dev env, nor any real mail would be sent to anyone.

Update required in [wiki](https://github.com/glittergallery/GlitterGallery/wiki/Usage-Instructions)
* execute: `gem install mailcatcher`
* Don't add it to gemfile. just run `mailcather` to start it
* Go to "127.0.0.1:1080" to see the mails. See [Details](http://mailcatcher.me/)

Solves: #232 

~~**#in_progress**~~ **#Needs_Feedback**


config/initializers/0_glitterconfig.rb
```diff
@@ -27,7 +27,13 @@
...
 # MAIL_AUTHENTICATION=plain
 Glitter::Application.config.action_mailer.delivery_method = :smtp
-Glitter::Application.config.action_mailer.smtp_settings={
+if Rails.env.development?
+	Glitter::Application.config.repo_dir="public/data"
+elsif Rails.env.test?
+	Glitter::Application.config.repo_dir="public/testdata"
+elsif Rails.env.production?
+	Glitter::Application.config.repo_dir="ENV['OPENSHIFT_DATA_DIR']/data"
+    Glitter::Application.config.action_mailer.smtp_settings={
```
rohitpaulk: I can see an indentation mismatch in these lines. It's possible that the indentation wasn't correctly used before this PR, but this is a good time to fix it. We use 2 spaces (not tabs) for indentation.

rohitpaulk:
> Had to change "project" in comment controller to instance variable for view to pick it up.

Please add a test case supporting this, so that the error doesn't happen in the future.

sonalkr132:
Let me know if it needs any changes. 
#First_test_ever_written

spec/controllers/comments_controller_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+
+describe CommentsController, type: :controller do
+	describe "POST #create" do
```
Alwahsh: It's really hard to read with such indentation. Can you please fix it?

sonalkr132: rookie error: mixing tabs and spaces for indentation!

sarupbanskota:
Thanks @sonalkr132. Tests are involved, so we'll wait for @rohitpaulk to have a look. In the meanwhile, I noticed some indentation issues thrown around - please fix them.

sonalkr132:
Sorry, I was mixing tabs and spaces for indentation. so everything looked fine in my text-editor. 
I had set tab_space =2 and translate_tab_to_space = true. I hope I should not have any more issue?

spec/factories/comments.rb
```diff
@@ -0,0 +1,4 @@
+FactoryGirl.define do
```
rohitpaulk: If you aren't using this factory, why define it?

sonalkr132: I thought it would give some error otherwise. 
hey.. we will use it someday?

spec/controllers/comments_controller_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+
+describe CommentsController, type: :controller do
+  describe "POST #create" do
+    before do
+      @project = FactoryGirl.create(:project)
+      sign_in(@project.user)
+    end
+    it 'Response code is 200' do
+      xhr :post, :create, comment: { polycomment_type: "project", polycomment_id: 3, issue: false, project_name: @project.name, body: "heads are cool" }
```
rohitpaulk: I don't know where this was broken, but looks like it's completely messed up. 

The whole point of `polycomment_type` and `polycomment_id` is that the former should refer to a class, and the latter should refer to the `id` of an object of that class. So in this case, `polycomment_id` should be set to a `project`'s id. Looks like we're using project_name here instead, to find the project - 

```
Project.find_by(name: params[:comment][:project_name])
```

We should also be checking `polycomment_id` to make sure that the object being commented on exists. This'll help to delete all dependent comments once their parent is deleted. 

Steps to do here - 

- Restore the use of `polycomment_id` and `polycomment_type`. 
- Remove `project_name`. 
- Create tests for an invalid `polycomment_id`, make sure a comment object is not saved to the database. 
- In this test, don't only check for a `200`, also make sure that the comment has been created in the database.

Extras: 
- Rename `polycomment` to `commentable` and use Rail's polymorphic associations. This way, we'll be able to access the object being commented on as `comment.commentable`. It'll automatically fetch a project or issue, whichever is appropriate.

sonalkr132: Should I make keep submitting changes to current PR or should I start a new one? what do you think about this: https://github.com/jackdempsey/acts_as_commentable

rohitpaulk: > what do you think about this: https://github.com/jackdempsey/acts_as_commentable

I don't think we need it, but if you want to use it - go ahead.

> Should I make keep submitting changes to current PR or should I start a new one?

Your choice :)

sonalkr132: I think we are eventually going to implement glitterposts too. acts_as_commentable will be useful.
I going to keep pushing to same PR. You can mark it in progress?

Since we are going in details and why not add [acts-as-taggable-on](https://github.com/mbleigh/acts-as-taggable-on) as well?

sonalkr132:
Any suggestion about how do I tell new comment form that it is commentable for projects? Basically how do i set commentable_type? I don't want to hardcode it like this:
```Ruby
%section.newcomment
        = render partial: "comments/new", locals: {type: 'project', id: @project.id.to_s, ajax: @ajax, comment: @comment}
```
Can't form be smart enough to pick commentable_type, commentable_id on it's own?

sonalkr132:
Also, which models are going to have polymorphic association with comments? Are glitterpost and projects only two?

rohitpaulk:
We aren't using glitterposts anymore. Projects, and issues are all I can think of right now - not sure if there are any others.

rohitpaulk:
> Can't form be smart enough to pick commentable_type, commentable_id on it's own?

Yh, it should be. Can you confirm whether `commentable: @project` works?

sonalkr132:
I am not sure how `commentable: @project` suppose to work.
In create method of comments we have to:
- find current project so that we can deliver notifications
- find commendable object so that we associate comments with it

only way I see is that we have to pass:
- project_id
- commentable_type
- commentable_id

as hidden fields.

Only improvement we can possibly do is replace:
```Ruby
@comments = Comment.where(polycomment_type: params[:comment][:polycomment_type],
                              polycomment_id: params[:comment][:polycomment_id])
```
with 
```Ruby
@comments = @commentable.comments
```
and may be find projects by their id instead of name.

sonalkr132:
My last commit is possibly the best solution I know.

Besides *issues* and *comments*, *file, blob and commit* are also commentable. They don't have a model, so I don't think we can use do things defining relationship on model and then later finding the object on which we have to comment (as there is no object)

EmilyDirsh:
@sonalkr132 IIRC, that's why the comments were implemented in this way in the first place. We want to be able to comment on files and on the history, and possibly other git-related objects we don't have models for. So, we were using other means of tracking which comments belong to which things. My advice would be to revert your last two commits to take out commentable, and work on fixing the comments using the polycomment model we had before.

sonalkr132:
Yeah, What we had was unconventional but thats probably what we need. #185 
>we aren't making use of the rails syntactic sugar when working with comments in controllers.

We can't unless we have a model for all the commentable objects.

@rohitpaulk I have made changes to create method of comments. It was way too long. Also, I have defined a new method to create notification as we are using it at quite few places. 
We do need to make it more readable, action = 0, 1.. 5 doesn't make much sense. TBH, I still don't have much clue what they are except that they are different for different actions. 
I have tired to add test which fails if commentable object is not found, but it excludes *files, commits and blob*.

rohitpaulk:
> We can't unless we have a model for all the commentable objects.

True, I should've thought of that. I've closed #185.

rohitpaulk:
> We do need to make it more readable, action = 0, 1.. 5 doesn't make much sense.

I'm all for making it more readable. If you want to know what they mean - there's a long comment at https://github.com/glittergallery/GlitterGallery/blob/master/app/models/notification.rb#L8

app/controllers/application_controller.rb
```diff
@@ -98,4 +98,15 @@ def after_sign_in_path_for(resource)
...
+  #used to deliver notification on various events
+  def notify_users(action, object_type, object_id, victims)
+    Notification.create(
+          :actor => current_user,
+          :action => action, 
+          :object_type => object_type,
+          :object_id => object_id,
+          :victims => victims
+        )    
+  end
```
sonalkr132: @rohitpaulk May be we can move logic of deciding which action number to use in this function as well? we can pass action argument as strings like "project_comment", "project_follow", "project_fork", "follow_user" etc.
So function call would look something like:
```Ruby
notify_users "project_comment", 1, @comment.id, victims
```

app/controllers/application_controller.rb
```diff
@@ -99,4 +99,14 @@ def after_sign_in_path_for(_)
...
+  def notify_users(action, object_type, object_id, victims)
+    Notification.create(
+          actor: current_user,
+          action: action,
+          object_type: object_type,
+          object_id: object_id,
+          victims: victims
+        )
+  end
 end
```
sonalkr132: Can we pass action as readable strings like "project_comment", "project_follow", "user_follow" etc? We can pass those strings as action argument and then use it to decide which action number do we need.
Would it be helpful in making actions more readable?

app/views/projects/projectcommit.html.haml
```diff
@@ -18,8 +18,8 @@
...
-      = render partial: "comments/new", locals: {type: 'commit', id: @id.to_s, ajax: @ajax, comment: @comment}
-
-    %ul.comments
-      = render partial: "comments/comment", collection: @comments
\ No newline at end of file
+    %section.comments
+      %section.showcomments
+        = render partial: "comments/comment", collection: @comments.reverse
+      %section.newcomment
+        = render partial: "comments/new", locals: {type: 'commit', id: @id.to_s, ajax: @ajax, comment: @comment}
```
sonalkr132: create.js of comment uses `.comments .showcomments

app/controllers/comments_controller.rb
```diff
@@ -8,43 +8,26 @@ def new
...
-      flash[:alert] = 'Something went wrong, try reposting your comment.'
-    end
-  end
-
-  def notify_followers(project, comment)
-    if comment.polycomment_type == 'project'
-      action = 0
-    elsif comment.polycomment_type == 'issue'
-      action = 5
+      redirect_to root_url
```
rohitpaulk: This is a bad request, I'm not comfortable with swallowing the error and sending a redirect. I think we should send a 4xx request code.. 404 seems apt.

sonalkr132: Okay.. that is easy.

app/controllers/application_controller.rb
rohitpaulk: > Can we pass action as readable strings like "project_comment", "project_follow", "user_follow" etc? We can pass those strings as action argument and then use it to decide which action number do we need.
Would it be helpful in making actions more readable?

Definitely, it would. I'd suggest storing the keys in a hash in the model [like this](https://github.com/glittergallery/GlitterGallery/blob/master/app/models/issue.rb#L21-L27).

> Should we just fix it for now by adding default case in objectname, since we don't even have view for files and blobs or should we completely refactor actions, since we also need to make it readable?

Completely refactor. If we're going to support comments/actions for files/commits, we should find a way to show the user a proper string that identifies the object - It's okay if it doesn't have a view now, we can implement that later.

sonalkr132:
Hey, who should receive the notification if a user forks a projects? Just the owner of repo or all the followers? I don't think anyone gets notification when someone forks a project on github.

sonalkr132:
This would be third time I have met a dead end because we can't access objects of rugged (or can we?). 
First was when I was trying to make objects of rugged `commentable`, second would be when I was trying to move DB related functions to model and the latest one is:
I am unable to deliver notification when someone comments on a commit. I need to find the commit object so that I can point user to that when I deliver the email. 
Until now `polycommentable_type` and `polycommentable_id` was enough to find the object (like project, comment, user etc) associated with comment/action but I don't think we can find `commit` object just by knowing those two parameters.

I tried to look into code of **GitLab** but they don't seem to be using Rugged. I can think of work around to deal with the problem I are having(I would need to find the project first and then the commit) but this 'no model scenario' need to be deal for good. Does  anyone know how?

rohitpaulk:
> I can think of work around to deal with the problem I are having(I would need to find the project first and then the commit) but this 'no model scenario' need to be deal for good. Does anyone know how?

I get the problem, would adding `project_id` as an attribute solve the problem? If you knew the project, can you identify/link to commits and all that?

sonalkr132:
Yes, if I know the project I can find the commit using `commit_id`. I project model we have:
```Ruby
def commit(id = nil)
    get_object :commit, id
 end
```
So we won't need to write any extra code either. However, just telling the notification model about the project object is a challenge in itself. What I can do is add function in comment controller whose purpose is just to pass `project_name` to notification model. I found it on [stackoverflow](http://codereview.stackexchange.com/questions/6763/should-i-pass-my-params-from-controller-to-the-model) (another [link](http://stackoverflow.com/questions/10237388/how-do-you-pass-data-from-a-controller-to-a-model-with-ruby-on-rails)). 
As you can tell, it is just adding complexity to an already complex commenting system and violates MVC. 
Another idea which might work is we declare our own commit,file and blob model and make an object of our model when we create object of rugged. We can establish a relationship between them (our model and rugged object) by using `id` of the rugged objects. This would also allow us to use the syntactic sugar like `commentable` as we would define it on our model. We would also be able to mode DB related operation of notification to model of these. And of course it would allow us to find commit object without knowing about project. 

TBH, I am not even sure if something like that can work. That does some of our problems but there has to be a better way!

sonalkr132:
@rohitpaulk hey.. which path should I follow? Should use a method just to tell notification model about project object or should I try out defining my own commit model and linking that with rugged?

rohitpaulk:
> should I try out defining my own commit model and linking that with rugged

When you say 'commit model', do you mean that it's persisted in the database? 

All we need is a method to link to a certain commit in a project. Wouldn't the commit's SHA-1 hash do?

Alwahsh:
@rohitpaulk have a look at the first paragraph [here](https://fedoraproject.org/wiki/GSOC_2015/Student_Application_aaboouelhamayed#Better_commenting_system). I believe that's what he wants to do.

sonalkr132:
> All we need is a method to link to a certain commit in a project

I know we can do it and it's all we need now.
> do you mean that it's persisted in the database?
yes! I think it will solve this problem and many others (at least thats what I think). If we have a model then we can move saving of notification in DB to model of our model (which we can't right now), we can define a `commentable` relationship in our model. and remove all the complexity we have right now in relating comments to objects. 
I know it sound kinda stupid as thats exactly why we have rugged for. But anyway, what do you think? 

@Alwahsh tableless models? I want to have model in have tables. We can add information to our table which we can't with rugged. For ex: define relationship with other models, perform validations, define our own functions. I don't know. Does it sound stupid?

Alwahsh:
> @Alwahsh tableless models? I want to have model in have tables. We can add information to our table which we can't with rugged. For ex: define relationship with other models, perform validations, define our own functions. I don't know. Does it sound stupid?

I think that wouldn't be efficient... You're duplicating information which is not a good idea.

sonalkr132:
Yeah that's what I think too. There is a tradeoff and I couldn't decide if it was worth it. Hence I asked.. 
Anyway, does tableless model help here? Tbh! I don't know how do they work. Can it help in the problem I am having(finding commit object in notification model). 

May be you can take it from here? So far, I have fixed the issue of ajax comments, make sure comments gets added only if the commentable object exists, added tests, make a function to deliver notification, use rebadle string for action and added fb like comment box style.

sonalkr132:
@Alwahsh if you are busy then may be you can just explain it a little and I can do it?

Alwahsh:
I think tableless models are a good solution. I'm not so experienced with them either. I just have a big picture of how things would look like but I'm sure some problems would show up. It's a challenging part. I don't think it's something to work on at the current time. I'm more interested in refactoring what we have first and make everything work correctly and then introduce these changes later :)

sonalkr132:
>  I just have a big picture of how things would look like but I'm sure some problems would show up. It's a challenging part. I don't think it's something to work on at the current time.

@rohitpaulk should I work on work around (it may get messy)? Or should we skip the part where we test that we are adding comments only when `commit` object exits and revisit it again? 

> All we need is a method to link to a certain commit in a project. Wouldn't the commit's SHA-1 hash do?

We can't form link which looks like `http://localhost:3000/user_name/project_name/commit/f4f18445cfb928e02535768ab1ca21be15202c1f` by just knowing the issue. We need to know the `project_name
>  I have fixed the issue of ajax comments, make sure comments gets added only if the commentable object exists, added tests, make a function to deliver notification, use readable string for action and added fb like comment box style.

I have added test for comments being made only if *issues* and *project* exist. Is these anything else in those changes which need to be tested?

rohitpaulk:
> We need to know the project_name but notification model knows only about action_id, 'action,object_type,object_id`.

Maybe we should provide the notification model with the project too, then?

sonalkr132:
Thats the work around I am talking about. But think of notification related to user follow and other user activities, there our `project_name/id` would be useless. It doesn't feel like a natural solution. What alwash suggested can be a solution and as per him it would require more research and time. 

btw.. gitlab uses [gitlab_git](https://github.com/gitlabhq/gitlab_git) and the gem uses rugged. So I was wrong, Gitlab uses rugged. If we snoop around their code, I am sure we can figure out how they are doing it.

sonalkr132:
It is annoying to work on other features when comments are not working, I guess thats why @Alwahsh  was completed to fix it :stuck_out_tongue:  I would like to remove `in progress` from here. Does this PR need any changes? We should go for a stable solution and I guess that would be a new PR.

Besides if this gets merged then I will be only 26 commits from overtaking you :D So there is that! I know some of the commits in this PR are just merge from upstream but I hope I am handling that better now in my other PRs.

sarupbanskota:
@sonalkr132 I'm merging this, waiting on your summary for this issue and we can take it up on another issue eventually.

sonalkr132:
Things left to be on comments done even after this merge:
* Ensure file, blob, PR, commit are commentable (some are broken)
* Check if commentable object( file, blob, PR and commit) exists before creating comment
* Threading of comments
* Ping back support for IRC

A solution which allows use of syntactic sugar (commentable) would make a lot things easier.
Possible solution : Tableless models
Research: Gitlab commenting system

PR #255, opened 2015-03-12
Upgrade to rails 4.1

sonalkr132:
* changing "http://rubygems.org" to "https://rubygems.org"  fixed "Bundler::MarshalError TypeError: incompatible marshal file format" error during bundle install
* rails 4.1 has better way to handling secret keys for different environments
* rails 4.1 doesn't support default_scope without block, hence change in comment.rb. Details on [stackoverflow](http://stackoverflow.com/questions/24163571/default-scope-activerecord-4-1-block-usage-for-ordering#answer-24164710)  

I know I have sent this PR earlier too but they were just messed up. 
#trying to follow suggestions from emily


config/secrets.yml
```diff
@@ -0,0 +1,35 @@
...
+  <<: *defaults
+
+test:
+  <<: *defaults
+
+# Do not keep production secrets in the repository,
+# instead read values from the environment.
+production:
+  secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
+  devise_secret_key: <%= ENV["SECRET_KEY_BASE"] %>
```
rohitpaulk: Everything looks good to me, except here - Should these 2 keys both be the same?

rohitpaulk:
Other that one comment, good to merge.

config/secrets.yml
sonalkr132: It's not going to cause any error or security risk. I just didn't want to introduce anymore environment variables. Devise requires one to send verification token sent in email during confirmation/reset password of account and rails uses it to make signature for sessions.    

Should I use two?

rohitpaulk: > It's not going to cause any error or security risk.

I'm not sure about that. Leaking the devise secret key could be far less dangerous that leaking the rails secret_key, right? 

Unless you have a strong reason - to be safe, let's use two different env vars.

sonalkr132: Once leaked both of them will have same consequence, in case of devise hacker will be able to send request to reset password, actual mail doesn't need to be sent to user email, he just needs the reset link (hence gain control of account) and in case of rails he will be able to forge any user session(hence again gain control of account). 

But yeah.. lets use two different ones.

sonalkr132:
Wouldn't it better if we move those two variables to "0_glitterconfig.rb"?

PR #257, opened 2015-03-12
Fixed Avatar Loading (Local)

mecyborg:
Added the public github image link with check. As for gravatar to load, the image source be public.


app/helpers/application_helper.rb
```diff
@@ -23,7 +23,7 @@ def avatar(email = nil)
     else
       gravatar_id = Digest::MD5.hexdigest(email.downcase)
       gravatar_size = Rails.application.config.gravatar_size
-      tag :image, src: "http://gravatar.com/avatar/#{gravatar_id}.png?s=#{gravatar_size}&d=#{CGI.escape(root_url + default)}"
+      tag :image, src: "http://gravatar.com/avatar/#{gravatar_id}.png?s=#{gravatar_size}&d=#{CGI.escape(default)}"
```
rohitpaulk: This will work for development, but not for production. `default_avatar` shouldn't be relative, it should include the root_url too.

config/initializers/0_glitterconfig.rb
```diff
@@ -7,7 +7,11 @@
 
 # (gr)Avatar options
 ## path to default avatar - should start with slash
+if Rails.env.development?
+Glitter::Application.config.default_avatar='https://raw.githubusercontent.com/glittergallery/GlitterGallery/master/public/happypanda.png'
```
rohitpaulk: Please fix the indentation here - use 2 spaces to indent a block.

app/helpers/application_helper.rb
mecyborg: Adding a check for Rails development environment.

config/initializers/0_glitterconfig.rb
mecyborg: Fixed. Thanks :)

mecyborg:
@rohitpaulk Please verify.

app/helpers/application_helper.rb
```diff
@@ -23,7 +23,11 @@ def avatar(email = nil)
     else
       gravatar_id = Digest::MD5.hexdigest(email.downcase)
       gravatar_size = Rails.application.config.gravatar_size
-      tag :image, src: "http://gravatar.com/avatar/#{gravatar_id}.png?s=#{gravatar_size}&d=#{CGI.escape(root_url + default)}"
+      if Rails.env.development?
+        tag :image, src: "http://gravatar.com/avatar/#{gravatar_id}.png?s=#{gravatar_size}&d=#{CGI.escape(default)}"
+      else
+        tag :image, src: "http://gravatar.com/avatar/#{gravatar_id}.png?s=#{gravatar_size}&d=#{CGI.escape(root_url + default)}"
+      end
```
EmilyDirsh: This'll work, but just some constructive criticism: this isn't very DRY - I'd suggest using the environment check to set a variable (`gravatar_default` or something similar) and use that variable in the image src. That'll cause less repetition, and make the URL just a bit easier to read.

mecyborg: Yeah. That's a better way to do this. Thanks.

mecyborg:
@EmilyDirsh @rohitpaulk Please check the new commit. Thanks.

app/helpers/application_helper.rb
```diff
@@ -23,7 +23,12 @@ def avatar(email = nil)
     else
       gravatar_id = Digest::MD5.hexdigest(email.downcase)
       gravatar_size = Rails.application.config.gravatar_size
-      tag :image, src: "http://gravatar.com/avatar/#{gravatar_id}.png?s=#{gravatar_size}&d=#{CGI.escape(root_url + default)}"
+      if Rails.env.development?
+        gravatar_default = "http://gravatar.com/avatar/#{gravatar_id}.png?s=#{gravatar_size}&d=#{CGI.escape(default)}"
+      else
+        gravatar_default = "http://gravatar.com/avatar/#{gravatar_id}.png?s=#{gravatar_size}&d=#{CGI.escape(root_url + default)}"
+      end
```
rohitpaulk: The point of keeping things DRY, is to not repeat stuff unless you have to. We don't need to repeat `http://gravatar.com/avatar/#{gravatar_id}.png?s=#{gravatar_size}&d=#{CGI.escape(` here.

mecyborg: Please check the updated code.

app/helpers/application_helper.rb
```diff
@@ -23,7 +23,12 @@ def avatar(email = nil)
     else
       gravatar_id = Digest::MD5.hexdigest(email.downcase)
       gravatar_size = Rails.application.config.gravatar_size
-      tag :image, src: "http://gravatar.com/avatar/#{gravatar_id}.png?s=#{gravatar_size}&d=#{CGI.escape(root_url + default)}"
+      if Rails.env.development?
```
Alwahsh: Is it really necessary to do this check on every request. IMHO, we should keep that decision only in initialization and use the variable directly here.
i.e: make gravatar_default defined in 0_glitterconfig.rb
what do you think @rohitpaulk ?

rohitpaulk: Do we have access to `root_url` in an initializer?

rohitpaulk:
@Alwahsh - I'm going to merge this for now, if you can find an elegant way to avoid that check - you know what to do :)

Alwahsh:
@rohitpaulk  yeah you're right. I actually feel there's something beter to be done but I couldn't find it yet :)

app/helpers/application_helper.rb
mecyborg: @Alwahsh I think its better keep this as a local check rather than defining it globally.

sonalkr132:
Even on openshift, it uses the default image(happypanda.png), when it should use avatar. Anyone working on it should look into that too.

Issue #239, opened 2015-03-09
[RFE] Container support


EmilyDirsh:
tl;dr: Containers are awesome, and GlitterGallery should work in containers.

Containers are becoming increasingly important to IT infrastructures, and therefore to developers and ops folks deploying applications using containers. OpenShift v3 is in beta now and is completely containerized. Being able to run Glitter Gallery in a docker container will allow us a really easy way to distribute GG, and for our contributors and users, easier set up of development environments and easier deployment.

Check out the interactive tutorial for the basics: https://www.docker.com/tryit/

EmilyDirsh:
@sarupbanskota I know you've started working on this some. Can you track your progress in this issue? If you want to continue working on it, assign it to yourself. If you don't have time, let's just leave it open for whoever wants to pick it up.

sarupbanskota:
I'll leave it open. I have inconsistent bursts of time available, don't want to be a blocker :)

ChenLiZhan:
I have some experience on Docker, can we discuss more detail :)

ChenLiZhan:
I think you want GG becomes a dockerize application! Is there anything that might I need to aware?

EmilyDirsh:
#259 is addressing this

PR #259, opened 2015-03-12
Run GlitterGallery in a container

Alwahsh:
I tried to learn about Docker and managed to run GlitterGallery in a container for development through this docker file.

Instructions:
-[install docker](https://docs.docker.com/installation/).
-cd to GG directory
-run `docker build -t container_name .`
-run `docker run -p 3000:3000 container_name`
-navigate to "localhost:3000" and enjoy :)

To make this run in production. I think we need to setup another container with mysql and connect them. Also we might want to use Passenger.

I just want to make sure that I'm working on the right track for handling #239 .


Alwahsh:
@sarupbanskota, @EmilyDirsh  I hope you can help me with this :)

Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM ubuntu
```
rohitpaulk: It's good to specify a version number, usually the LTS version - which would be 14.04.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
```
rohitpaulk: Add yourself as a maintainer here :)

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN curl -sSL https://get.rvm.io | bash -s stable
+RUN /bin/bash -c -l 'rvm requirements'
+RUN /bin/bash -c -l 'rvm install ruby-2.2.0'
+RUN /bin/bash -c -l 'rvm use 2.2.0'
+RUN /bin/bash -c -l 'gem install bundler --no-ri --no-rdoc'
+
+
+# Create a directory for the app.
+RUN mkdir -p /glittergallery
+WORKDIR /glittergallery
```
rohitpaulk: We generally place app files either in the home directory, or in `/srv/appname/`. I've only seen `/srv/` and `/var/www` been used in Dockerfiles though.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Allow reaching via port 3000.
+EXPOSE 3000
+
+# Install required gems.
+ADD Gemfile /glittergallery/Gemfile
```
rohitpaulk: If you're copying the whole application anyway, why do you have to copy these two files individually beforehand?

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+ADD Gemfile.lock /glittergallery/Gemfile.lock
+RUN /bin/bash -c -l 'bundle install'
```
rohitpaulk: Hmm, I haven't seen this `/bin/bash -c -l` usage anywhere - I'm curious to know what it does. Can't you just run the commands directly?

rohitpaulk:
> docker build -t container_name .

You're building an image here, not a container. I'd suggest replacing `container_name` with `image_name`, or better suited for our case - just `glittergallery`. 

> docker run -p 3000:3000 image_name

That's good to run GlitterGallery in a setup where you just want to set it up and forget about it. If you're actively developing, it's better to mount a volume from the docker host into the container - here's how that'd look: 

`docker run -d -v $PWD:/glittergallery -p 3000:3000 image_name`

This way, all changes made to files will reflect in the container. I'm a bit doubtful about how this will work with bundler (if gems are installed in the /glittergallery folder, they'll be overwritter). If that is the case, we'll have to configure bundler to install gems elsewhere. 

One more advantage of this is - you'd retain the development database even if you rebuild the container.

Dockerfile
Alwahsh: Docker keeps caches after each step, copying the whole thing would make the cache size bigger if it fails at bundling for example. It's advised to keep that part of copying the actual app to the end of the file :)

rohitpaulk: Aah, didn't think of that. Definitely makes sense.

Dockerfile
Alwahsh: Running the commands directly doesn't work. I don't entirely get the point of doing these :$ but in Ubuntu with GNOME shell there was some setting to edit to make rvm work. I think adding /bin/bash -c -l before the command just does that setting. Not entirely sure though :)

Alwahsh:
@rohitpaulk Thank you very much for the detailed feedback :)

I posted this just as a starting point. 
`docker build -t container_name .` was just for explaining what that argument is. When I ran it locally, I named it glittergallery indeed :hand: :) .

>That's good to run GlitterGallery in a setup where you just want to set it up and forget about it.

Yeah, or in other words "production" :)
I don't actually develop with using Docker. I just wanted to run it in my environment to get used to it and know what's needed to run it in production and to get some feedback here before making a real production version Dockerfile where I'd set the port to 80 and the ENV to production and everything :).

Thanks again. I really appreciate your feedback

Dockerfile
Alwahsh: yeah I agree.

Dockerfile
Alwahsh: Alright :)

Dockerfile
Alwahsh: Sure. That should definitely done in production :)

EmilyDirsh:
Feel free to continue to work on this Ubuntu based image, if that's what you prefer working with. I'd want our promoted "official" image to be Fedora based, given that Glitter Gallery originated with the Fedora design team :)

Alwahsh:
@EmilyDirsh I'm planning to change it to Fedora :)... Funny thing here is that I was actually downloading Fedora for my own machine during the past few days(internet here is not the fastest :cry: ) I was facing problems on my old system getting the image to run(discovered at the end that I had no space left in my root.) Just today as I removed my old system and switched to Fedora that I got this Ubuntu based image to work :).

Dockerfile
rohitpaulk: @Alwahsh - `ENV DEBIAN_FRONTEND noninteractive`. :)

Alwahsh: @rohitpaulk  that doesn't seem to fix it - if that's what you meant by that line. - :)

rohitpaulk: Hmm, weird. I'll try this out on my local in some time and let you know.

Alwahsh:
@rohitpaulk I've applied the suggestions you made. I could run the app in development, make modifications and see them directly. I think next step is to run a container with guard running :)

@EmilyDirsh now using Fedora :)

sarupbanskota:
Do we really need rvm in the container? The only reason I can think of wanting to switch ruby versions within a container I'm going to use for GG development is to check if some bug is introduced in a project because of a change we introduced when we were on a previous version. That's something that'll have to reflect on the Dockerfile anyway.

Am I missing something?

Alwahsh:
@sarupbanskota yeah I was thinking actually about that but found out that the official available ruby packages is not always the last one. The easiest alternative is RVM. If you think I should use another way to get the latest ruby with specifying the version, please let me know it :).

EmilyDirsh:
I'm thinking we should have the Dockerfile in a separate repo. That's standard from what I've seen of other docker repos.

Alwahsh:
@EmilyDirsh well, I was thinking of including that file in the repo here and adding something in the instructions in the [Get Involved](https://github.com/glittergallery/GlitterGallery/wiki/Get-Involved!) page to tell contributors they can optionally develop that way since this dockerfile is now for development. For the production one(probably won't be a single file), it might be better to move them into another repo. 
I'm not sure though. If this file looks good and you want to host it in another repo, please do :)

EmilyDirsh:
@Alwahsh that's a good point, but I still think it's best to keep it in it's own repo. If we want to include it in this repo at some point, then we can do that through a [submodule](http://www.git-scm.com/book/en/v2/Git-Tools-Submodules). I'll make it a todo for myself to create the repo.

EmilyDirsh:
Addresses #239

PR #265, opened 2015-03-15
move project header to partial

sonalkr132:
210 line DRY. Win!

### DIFF
--- a/app/views/projects/user_show.html.haml
+++ b/app/views/projects/user_show.html.haml
@@ -29,8 +29,7 @@
             %p
               %img{:src => "/usercover.jpg", :title => "Project Cover"}
               = link_to project.name, project.urlbase
-              - @project = project
-              = render 'forked_from', :project => @project
+              = render 'forked_from', project: project
               %br
               Last updated #{distance_of_time_in_words_to_now project.updated_at} ago
   - else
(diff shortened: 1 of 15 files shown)


sonalkr132:
Any idea what to do with this passing of local to fork_from partial? At first I felt it was unnecessary since instance variable of project was already available. 
But as you can see, it failed test, as there has been cases where it is called over project as local variable. In this case we can do:
```Ruby
- @project = project
= render 'forked_from', :project => @project
```
Looks kinda unconventional. 
So either we can do this(change local to instance when needed) and be unconventional or we can always pass project as local variable to all the partial calls of fork relation when instance variable of the same is already available.

app/views/projects/user_show.html.haml
```diff
@@ -29,7 +29,8 @@
             %p
               %img{:src => "/usercover.jpg", :title => "Project Cover"}
               = link_to project.name, project.urlbase
-              = render partial: 'forked_from', locals: {project: project}
+              - @project = project
+              = render 'forked_from', :project => @project
```
EmilyDirsh: IMO, it's not ideal to define instance variables in views. They should be defined in controllers. Also, if you're using an instance variable in the partial, you don't _also_ need to pass it as a local. I'd prefer we just pass it as a local variable every time.

sonalkr132: In **most** case we have the instance variable available, but there are a few where we don't. Like the case where we are iterating through the list of projects:
```Ruby
%div
   - @projects.each do |project|
     %div
       %p
```
Pipes don't support instance variable. 
Now, if we always pass it as local then first it is passed to project_header and then it project_header passes it to  fork_from, seems kinda ehh.. idk! Doesn't feel right.
But at least it will maintain uniformity through out the code.

PR #267, opened 2015-03-18
resolved #151

### LINKED ISSUE #151: Improve Issue ID lookup
Currently, the Issue subIDs are found by searching an array of the project's issues IDs, sorting and then finding the position of a certain issue. It's likely to be faster if we store the issue subID in our database. 

[Here](http://rohitpaulk.roon.io/title-undecided)'s a blogpost about this if you'd like to read more.

ghost:
I added the sub_id attribute on the issue model instead of calling a lot of queries to retrieve it from the project model.

### DIFF
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -1,7 +1,7 @@
 class Issue < ActiveRecord::Base
-	after_create :set_sub_id
+  after_create :set_sub_id
 
-	# This is to avoid conflict with the :type attribute
+  # This is to avoid conflict with the :type attribute
   self.inheritance_column = nil
 
   belongs_to :user
@@ -52,8 +52,8 @@ class Issue < ActiveRecord::Base
     Issue.find(project.issue_ids[sub_id.to_i - 1])
   end
 
-	def set_sub_id
-		current_id = project.issues.count
-		update_attribute(:sub_id, current_id)
-	end
+  def set_sub_id
+    current_id = project.issues.count
+    update_attribute(:sub_id, current_id)
+  end
 end
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -49,7 +49,7 @@ class Issue < ActiveRecord::Base
   end
 
   def set_sub_id
-    unless ub_id
+    unless sub_id
       sub_id = project.issues.count
       update_attribute(:sub_id, sub_id)
     end
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -8,6 +8,7 @@ class Issue < ActiveRecord::Base
   belongs_to :project
 
   validates_presence_of :title, :description, :user, :project, :type, :status
+  validates :sub_id, uniqueness: { scope: :project }
 
   # TODO: make a list of 5 most popular types of issues
   #       people can raise on design projects.
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -49,9 +49,8 @@ class Issue < ActiveRecord::Base
   end
 
   def set_sub_id
-    unless sub_id
+    return if sub_id
       sub_id = project.issues.count
       update_attribute(:sub_id, sub_id)
-    end
   end
 end
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -50,7 +50,7 @@ class Issue < ActiveRecord::Base
 
   def set_sub_id
     return if sub_id
-      sub_id = project.issues.count
+      sub_id = project.issues.count + 1
       update_attribute(:sub_id, sub_id)
   end
 end
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -49,7 +49,7 @@ class Issue < ActiveRecord::Base
   end
 
   def set_sub_id
-    if !sub_id
+    unless ub_id
       sub_id = project.issues.count
       update_attribute(:sub_id, sub_id)
     end
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -44,12 +44,8 @@ class Issue < ActiveRecord::Base
     File.join(project.urlbase, 'issue', sub_id.to_s)
   end
 
-  def sub_id
-    (project.issue_ids.index(id) + 1).to_i
-  end
-
   def self.find_from_project(project, sub_id)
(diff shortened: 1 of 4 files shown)


sarupbanskota:
@AhmedZaleh could you pull from master? @rohitpaulk this addresses https://github.com/glittergallery/GlitterGallery/issues/151, so I'm assigning you for review.

AhmedZaleh:
I rebased my commit, please review and merge.

app/models/issue.rb
```diff
@@ -49,4 +51,9 @@ def sub_id
   def self.find_from_project(project, sub_id)
     Issue.find(project.issue_ids[sub_id.to_i - 1])
   end
+
+	def set_sub_id
+		current_id = project.issues.count
+		update_attribute(:sub_id, current_id)
+	end
 end
```
sonalkr132: Hi, your rubocop is failing. We use two spaces for indentation. If you are going to use tabs please configure your editor so that it uses 2 spaces as tabs.

AhmedZaleh: I fixed that :)

rohitpaulk:
There's a lot more to do here. You've added a `sub_id`, but you aren't using it anywhere. I'll add a few comments inline pointing to the functions

app/models/issue.rb
```diff
@@ -49,4 +51,9 @@ def sub_id
   def self.find_from_project(project, sub_id)
     Issue.find(project.issue_ids[sub_id.to_i - 1])
```
rohitpaulk: This should be modified, and the `sub_id` method should be deleted.

app/models/issue.rb
```diff
@@ -49,4 +51,9 @@ def sub_id
...
   end
+
+  def set_sub_id
+    current_id = project.issues.count
```
rohitpaulk: I'd prefer that `sub_id`s start from 1 instead of 0.

AhmedZaleh: the sub_id always starts from 1, because it's only set to the number of issues in a project after an issue is created (sub_id = project.issues.count)

rohitpaulk: > The sub_id always starts from 1, because it's only set to the number of issues in a project after an issue is created (sub_id = project.issues.count)

Aah, I'm sorry I didn't notice that it was `after_create`, I assumed it'd be `before_save`. 

> bedsides, sub_id will always be unique inside a project (project.issues)

Aah, a bit of a catch there. Under high traffic, it's possible for two issues to end up with the same sub_id. Think of this - you and me try to create a new issue at the same time. If we're both served by different threads, and your issue is created **just** before my `after_create` callback, I'll end up with a `sub_id` that is one more that what it's supposed to be, and you'll end up with the same. 

Normal case: 

```
1. I create my issue
2. I fetch the issues count, update my sub_id
3. You create your issue
4. You fetch the issues count, update your sub_id
```

Failing case: 

```
1. I create my issue
2. You create your issue
3. I fetch the issues count, update my sub_id
4. You fetch the issues count, update your sub_id # Both of us get the same sub_id. 
```

Placing this is a before_save callback with validations in place should greatly reduce the chance of this happening. If we don't want this to happen *at all*, I guess we'll have to set unique index constraints right down in the DB.

rohitpaulk: By the way, what OS/terminal do you use? It looks great!

AhmedZaleh: > Placing this is a before_save callback with validations in place should greatly reduce the chance of this happening. If we don't want this to happen at all

I do get your point but I don't how would we use a before_save callback for this issue, wouldn't it change the sub_id every time an issue is updated i.e., saved ?

> By the way, what OS/terminal do you use? It looks great!

I use Gnome Terminal on Fedora 21 ;)

rohitpaulk: > I do get your point but I don't know how would we use a before_save callback for this issue. Wouldn't it change the sub_id every time an issue is updated i.e., saved ?

Only set it if it hasn't been set. (use `||=`)

> I use Gnome Terminal on Fedora 21 ;)

Cool, thanks :)

AhmedZaleh: how about this ?

```
class Issue < ActiveRecord::Base
  before_save :set_sub_id

  def set_sub_id
    return if sub_id
      sub_id = project.issues.count
      update_attribute(:sub_id, sub_id)
  end
```

app/models/issue.rb
```diff
@@ -42,11 +44,13 @@ def show_url
...
-    (project.issue_ids.index(id) + 1).to_i
+  def self.find_from_project(project, sub_id)
+    project.issues.find_by_sub_id(sub_id)
   end
 
-  def self.find_from_project(project, sub_id)
-    Issue.find(project.issue_ids[sub_id.to_i - 1])
+  def set_sub_id
+    return if sub_id
+      sub_id = project.issues.count
```
rohitpaulk: Since this is run before saving, we should set it equal to `project.issues.count + 1`

AhmedZaleh: Yes, that's right. Thanks a lot :)

rohitpaulk:
Please add validations too, we need to make sure that issue's aren't saved to the database without a `sub_id`. We'll need validations for both `presence` and `uniqueness`, and tests that check that they work.

AhmedZaleh:
I know I must validate the ```presence``` but how can I validate the ```uniqueness``` inside project.issues ? I mean project1.issues.first.sub_id = 1 and project2.issues.first.sub_id = 1 so I cannot use ```validate_uniqueness_of :sub_id``` that way. I think I can define a method to ensure sub_id uniqueness inside project.issues but what if I have duplicates what should I set the duplicate sub_id to ? Should I just increment it by 1 or there is another different way?

Alwahsh:
@AhmedZaleh you can use:
```ruby
validates :sub_id, uniqueness: { scope: :project }
```

AhmedZaleh:
@Alwahsh Thanks a lot :)

Alwahsh:
np :)

rohitpaulk:
This is still marked as `in progress`, we'll need tests before we can merge this in.

Alwahsh:
@rohitpaulk I've created some tests that would be enough to test this change I believe in #292... I'd recommend a squash of commits though :)

PR #268, opened 2015-03-18
Solves unprotected redirect warning on hakiri

sonalkr132:
Hakiri warnings seemed pretty valid. 
Adding :only_path => true wouldn't work because it needs hash table. Read [here](https://www.owasp.org/index.php/Ruby_on_Rails_Cheatsheet#Redirects_and_Forwards)
Following the same blog, I added a function which raises error if path to which we are redirecting doesn't matches the regular expression patterns

UPDATE:
Successfully used rebase :smirk:

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -108,5 +108,15 @@ class ApplicationController < ActionController::Base
   def after_sign_in_path_for(_)
     dashboard_path
   end
-
+ 
+  #raises error if match against regular expression patterns fails
+  def check_redirect(url)
+     begin
+       if path = URI.parse(url).path
+         redirect_to path
+       end
+     rescue URI::InvalidURIError
+       redirect_to '/'
+     end
+  end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -108,15 +108,11 @@ class ApplicationController < ActionController::Base
   def after_sign_in_path_for(_)
     dashboard_path
   end
- 
-  #raises error if match against regular expression patterns fails
+
+  # raises error if match against regular expression patterns fails
   def check_redirect(url)
-     begin
-       if path = URI.parse(url).path
-         redirect_to path
-       end
-     rescue URI::InvalidURIError
-       redirect_to '/'
-     end
+    redirect_to url if URI.parse(url).path
+    rescue URI::InvalidURIError
+      redirect_to '/'
   end
 end
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -40,7 +40,7 @@ class ProjectsController < ApplicationController
       redirect_to dashboard_path
     else
       flash[:error] = "You don't have permission for this command!"
-      redirect_to @project.urlbase
+      check_redirect @project.urlbase
     end
   end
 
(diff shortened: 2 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -109,4 +109,10 @@ def after_sign_in_path_for(_)
     dashboard_path
   end
 
+  # raises error if match against regular expression patterns fails
+  def check_redirect(url)
+    redirect_to url if URI.parse(url).path
+    rescue URI::InvalidURIError
+      redirect_to '/'
```
rohitpaulk: Shouldn't we add some sort of a flash message to the user here?

sonalkr132: `rescue` will be triggered only when someone has sent something is params which redirects to something other than our site. This doesn't solves the issue that user can't send something dubious as params, it only makes sure that those malicious links don't work, which in turn solves the issue that user won't send anything malicious as those efforts will be futile. I hope you understand.

Alwahsh:
Wouldn't this be resolved when we change all these redirects to the appropriate object_path method?

app/controllers/projects_controller.rb
```diff
@@ -343,6 +343,6 @@ def return_context
     render_404 && return if @project.blank?
     return unless @project.deleted?
     flash[:alert] = 'The project you requested had been deleted.'
-    redirect_to user_path(@user)
+    check_redirect "/#{@user.username}"
```
Alwahsh: This one doesn't give that warning and I think changing the others to be like it is a better solution than abandoning the _path links to hardcoded ones.

sonalkr132: I think it wouldn't give warning if we use rails to define our routes rather than our own methods like `urlbase`. You are working on routes, I hope you will figure it out? I guess this can wait until then.

Alwahsh: yeah it wouldn't...  I can't see a warning for this line you changed... If it's there and I didn't see it please point me to it... This part is all ready to be replaced... I have already defined the routes of the projects. Issues not yet merged but they're also ready.

sonalkr132: Hey, you can sign up for hakiri, follow your branch on your own repo and you will be able to see if you have any warnings on that branch. It is free for pubic repos. Don't forget to confirm if all the redirect issues are solved, we can close this one.

Alwahsh: yeah I know... I'm talking about the master branch here. I went through the warnings but couldn't find that line so I was wondering if it's there and I couldn't find it

sonalkr132: I am pretty sure warnings are only for case where we used `urlbase`

Alwahsh: yep ... That's what I'm saying and why I'm surprised at this change in that line.

sonalkr132: >  This part is all ready to be replaced... I have already defined the routes of the projects. Issues not yet merged but they're also ready.

can please confirm by following your branch on hakiri that changes you made solves the redirect warning?

Alwahsh: I did NOT do those changes... They're just ready to be done and don't worry... I'm pretty sure they solve it :)

sonalkr132:
using url rails way solved most of redirect warning, so we won't need this anymore.

Issue #249, opened 2015-03-11
Profile page with notification link


philippneugebauer:
Is that explicitly wanted that get directed to a new page and losing the context?

I was really confused after I have seen that and from the usability point that is not a good idea in my opinion

EmilyDirsh:
Could you add some details about this? What action does this happen on, what context is lost? Thanks!

philippneugebauer:
I have fixed it in my pull request.

I meant, that you lose the context when you click on notification in your profile area, while the head bar still appears at the user profile and the project menu

EmilyDirsh:
@philippneugebauer okay, that makes more sense. It's probably an oversight.

PR #253, opened 2015-03-11
Improved notification design

philippneugebauer:
Fixes #249

PR #271, opened 2015-03-20
Fixed notification show

mecyborg:
Fixed the notification view.
Added notification link in navigation.


Alwahsh:
This removes the need for #253 and closes #249

PR #272, opened 2015-03-20
missing except action

sonalkr132:
Sorry about this! It got messed up in merge..
That demo is run from a separate branch.


app/controllers/relationships_controller.rb
```diff
@@ -1,5 +1,5 @@
 class RelationshipsController < ApplicationController
-  before_action :authenticate_user!
+  before_action :authenticate_user!, except: [:follow]
```
rohitpaulk: Can you explain why you're doing this?

sonalkr132: Identify_user has the logic of redirecting if user is not signed in. But it won't be run unless we by-pass `authenticate user`. #197 suggests that we should redirect to login page instead of giving an unauthorized error.

I am gonna add a test in PR.

sonalkr132:
Okay.. I need help. I don't know to write test for this:
```Ruby
 def identify_user
     @user = User.find_by(username: params[:id])
     render js: "window.location = '/'" if @user.nil? || !user_signed_in?
   end
```

I tried this:
```Ruby
scenario "Follow when user is not logged in", js: true do
    sign_up_with("t@test.com","test1","secret12345")
    click_link "logout"
    visit "/test1"
    click_link "Follow test1"
    expect(current_path).to eql('/')
  end
```
Didn't work :stuck_out_tongue_closed_eyes:

rohitpaulk:
@sonalkr132 - Post the error please? Actually, integration tests *can* be kind of elusive - You could write a controller test first to make sure that you aren't missing anything. There should be a few examples in `specs/controllers` to help you out.

Alwahsh:
we don't have any driver that allows js testing yet

Alwahsh:
@sonalkr132 There seems to be something wrong... authenticate_user method does not actually return unauthorized error... It redirects to the login page. In your test, shouldn't you be expecting the current path to be `/users/sign_in` ?

sonalkr132:
I had added selenium-webdriver in my gemfile, so it works. My current test doesn't give any error but it does what it is expected.. it excutes `window.location = '/'` and thus open us my default browser.

@Alwahsh My demo doesn't give that error because it has 
```Ruby
 before_action :authenticate_user!, except: [:follow]
```
It was my PR where I forgot to restore it after merge. Also.. I am not directing to `/users/sign_in` but root which happens to be `/users/sign_in` as well.

Alwahsh:
I'm sorry but I can't understand completely... Which error are you referring to by "that error" ?

sonalkr132:
@Alwahsh unauthorized error

Alwahsh:
well... authenticate_user! does not give an unauthorized error as far as I know. Are you sure of this?

Alwahsh:
At least tell us, what's the error that you're having with the test?

sonalkr132:
@Alwahsh I never said it gave me any error. I just don't know to test
```Ruby
render js: "window.location = '/'" if @user.nil? || !user_signed_in?
```
Code of the test I have posted doesn't give any error. It just executes `window.location = '/'` and my default browser opens up.

sonalkr132:
>Actually, integration tests can be kind of elusive - You could write a controller test first to make sure that 
>you aren't missing anything

@rohitpaulk I wrote the controller test by they we exactly the same as ones written by @Alwahsh  in features.
for ex, I wrote:
```Ruby
context "User is signed in" do
    describe "follow and unfollow" do
      before do
        @user1 = FactoryGirl.create(:user)
        sign_in(@user1)
        @user2 = FactoryGirl.create(:user, email: "t@t.com", username: "tester")
      end

      it "allows user to follow other users" do
        expect {post :follow, :id => @user2.username}.to change {@user2.followers.count}.from(0).to(1)
      end
   end
end
```
and alwash has written:
```Ruby
scenario "Follows and unfollows others but not himself" do
    sign_up_with("t@test.com","test1","secret12345")
    click_link "logout"
    sign_up_with("t2@test.com","test2","secret12345")
    visit "/test1"
    expect(find('#social')).to have_link("Follow test1")
    expect(find('#social')).to have_content("0 Followers")
    expect(find('#social')).to have_content("0 Following")
    click_link "Follow test1"
    expect(find('#social')).to have_link("Unfollow test1")
    expect(find('#social')).to have_content("1 Followers")
    expect(find('#social')).to have_content("0 Following")
```
Should I push changes anyway? Aren't we performing same test?

rohitpaulk:
> @rohitpaulk I wrote the controller test but they we exactly the same as ones written by @Alwahsh in features.

Well, then those aren't the tests that you're looking for. You need a test that would fail unless the changes you've provided were made. I haven't looked in detail, but a quick skim tells me that the tests you just mentioned always have the user signed in - what you want to test for is when the user is **not** signed in.

sonalkr132:
@rohitpaulk There were no tests in controller for relationship. So I have two questions:
1. Tests I have written in the controller and ones alwash has written in feature are almost same. [code sample](https://github.com/glittergallery/GlitterGallery/pull/272#issuecomment-84143319) Should I push them here anyway?
2. How can I test this line:
```Ruby
render js: "window.location = '/'" if @user.nil? || !user_signed_in?
```
Since, normal way to check redirect won't work here.

rohitpaulk:
> Tests I have written in the controller and ones alwash has written in feature are almost same. code sample Should I push them here anyway?

Please read the comments carefully. That is **not** the test you are looking for. That test is for when the user is signed in - you want a test for when they are signed out. 

> How can I test this line:

If it's a feature spec with js testing enabled, then you can check for the current_path. If it's in a controller spec - check for a 200 response, and check that the content is `"window.location = '/'"`

sonalkr132:
> Please read the comments carefully. That is not the test you are looking for
I know! They were just samples and not the complete code. As you said:
> Actually, integration tests can be kind of elusive - You could write a controller test first to make sure that you aren't missing anything.
Hence I wrote them. Complete also includes the tests when user is not logged in. See my last commit.
>If it's in a controller spec - check for a 200 response, and check that the content is "window.location = '/'"
Thanks! I have changed my test accordingly. I didn't need to include driver for testing of ajax.

Alwahsh:
@sonalkr132 I ran the tests you added originall (not the latest ones) on master branch and they all passed:
```ruby
it "doesn't give unauhtorised error" do
  post :follow, :id => @user.username
  expect(response.response_code).not_to eq(401)
  expect(@user.followers.count).to be(0)
end
```
So it doesn't seem to be givin an unauthorized error on master right now. For your current test:
```ruby
it "redirects to root_path" do
  post :follow, :id => @user.username
  expect(response.response_code).to eq(200)
  expect(@user.followers.count).to be(0)
  response.body.should include("window.location = '/'")
end
```
it fails with the following:
expected: 200
got: 302

302 is actually the common method for performing redirection so I don't think what we have currently should be failing on this test. I understand that you'd want to reply with a different way to AJAX requests other than giving a redirect which would cause nothing to happen but I suggest that we modify the behavior of authenticate_user! to reply correctly to an AJAX request so that it would work well in the rest of the website.

sonalkr132:
> I ran the tests you added originally (not the latest ones) on master branch and they all passed:

This is response when you run master: 
screenshot from 2015-03-21 13 19 16

>it fails with the following:
expected: 200
got: 302

Thats is because your `relationship_controller.rb` looks like:
```Ruby
 before_action :authenticate_user!
```
and not ` before_action :authenticate_user!, except: [:follow]`

sonalkr132:
@Alwahsh  [stackoveflow](http://stackoverflow.com/questions/22944574/when-to-401-vs-when-to-302)

spec/factories/repos/dummy.git/refs/.gitignore
```diff
@@ -0,0 +1 @@
+/remotes/*
```
rohitpaulk: What is this doing here?

Alwahsh:
@sonalkr132 sorry but I can't find the relation between that link and what we're disucssing here... You don't like us giving a 302 for non logged in users? This is a very separate issue than what you're trying to do here and IMO giving a 302 is totally fine for HTML requests

spec/factories/repos/dummy.git/refs/.gitignore
sonalkr132: see [here](https://github.com/glittergallery/GlitterGallery/tree/master/spec/factories/repos/dummy.git/refs/remotes) @Alwahsh had added this for test. files like `bare/master` gets created here, after you run test locally. He has added a .gitignore file to ignore everything inside `bare/` but still the folder `bare` would get created and commited if we run test and commit. So we need to ignore everything inside `remote`

rohitpaulk:
> This is a very separate issue than what you're trying to do here and IMO giving a 302 is totally fine for HTML requests

I agree.

Okay, so I went back and read where this problem originated. Here's what I think we should be doing.

### HTML request
- **Signed in** - the usual. 
- **Not signed in** - redirect (302). 

### JS request
- **Signed in** - the usual, respond with code to update the counters on the page.
- **Not signed in** - Either render `"window.location = '/'"` to be executed as you've done, or we could raise a 401 - and have the JS callback handle the redirecting. I think the latter is a semantically better approach, but I can understand that the former would be simpler to implement.

rohitpaulk:
One more thing - If we're redirecting, we must remember the page that the user was on, and redirect to that once he's done logging in. I know we have mechanisms for this in many places in the app (`return_to`).

Alwahsh:
@rohitpaulk exactly... But we should handle the case of the JS request in the authenticate_user! method itself instead of defining another one and using except with authenticate_user! 
Am I right?

rohitpaulk:
> @rohitpaulk exactly... But we should handle the case of the JS request in the authenticate_user! method itself instead of defining another one and using except with authenticate_user! 
Am I right?

Yes, that would be a clean solution.

Alwahsh:
@sonalkr132 I hope you're convinced now :)
Do you want to implement this yourself or shall I do it?

rohitpaulk:
@sonalkr132 - I'll be on IRC for a while, feel free to pop in if you'd like to discuss this further :)

sonalkr132:
From link I posted:
>Sometimes when they make a request for a resource that requires authentication, I return 401 Unauthorized. Other times, I return 302 and redirect them to a login page.

Currently, master has `before_action :authenticate_user!`
 Right now master is using http request to handle follow request, so it redirects with 302, that is handled by devise. But we need ajax request! I was working on #197 and there it was suggested that we should redirect to login page.
>But we should handle the case of the JS request in the authenticate_user! method itself instead of defining another one and using except with authenticate_user! 

`Identify_user` was defined before I started working on it, it would be cleaner if we moved it to follow, but then follow would be lengthy. And since it was already defined there, I used it instead of moving it to follow.

rohitpaulk:
I'm closing this for now, we had a chat in IRC about this. 

> I think that after the discussion - we're at a consensus that using Devise's built-in 401 for authenticating, and handling redirection on the client would be the best way to go about it

PR #275, opened 2015-03-21
Handle js 401 response

Alwahsh:
Make 401 responses redirect to users/sign_in as discussed with @rohitpaulk on IRC :)


app/assets/javascripts/application.js
```diff
@@ -20,4 +20,10 @@
 //= require nprogress
 //= require nprogress-turbolinks
 //= require main
-//= require jquery.sidr.min
\ No newline at end of file
+//= require jquery.sidr.min
+
+$(document).ajaxError(function (e, xhr, settings) {
```
rohitpaulk: This would patch all Ajax requests, not only those created by Rails. So if a third party library that we integrate encounters a 401, we're going to end up taking the sure to the sign-in page.

spec/controllers/relationships_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  context "User is signed in" do
+    describe "follow and unfollow" do
+      before do
+        @user1 = FactoryGirl.create(:user)
+        sign_in(@user1)
+        @user2 = FactoryGirl.create(:user, email: "t@t.com", username: "tester")
+      end
+
+      it "allows user to follow other users" do
+        expect {post :follow, :id => @user2.username}.to change {@user2.followers.count}.from(0).to(1)
```
rohitpaulk: We should also test how the controller responds to JS requests vs HTML requests. We expect a redirect in one case, and a 200 in the other.

app/assets/javascripts/application.js
Alwahsh: @rohitpaulk 
Not sure I got that correctly, but see here:
[.ajaxError()](https://api.jquery.com/ajaxError/)
> Note: This handler is not called for cross-domain script and cross-domain JSONP requests.

spec/controllers/relationships_controller_spec.rb
Alwahsh: ok

app/assets/javascripts/application.js
rohitpaulk: > Note: This handler is not called for cross-domain script and cross-domain JSONP requests.

Oh, ok - that solves most of our problems. When I said "created by Rails", I meant created by [ujs](https://github.com/rails/jquery-ujs).

Looks like jquery-ujs does publish events for errors that we can use: 
https://github.com/rails/jquery-ujs/wiki/ajax.

It's just that - I'd like this to look less like magic. When we use `data-remote: true`, we're aware that we're using a function in Rails so it's okay to expect 401s to be handled for us. If someone ends up writing an AJAX call by hand, they shouldn't be surprised by the automatic 401 patch, they should be explicitly mentioning it in the error handler for that request. Two requests, one created by rails and one written by hand could end up calling the same error function, but we shouldn't assume that every request would need the 401 patch. Makes sense?

Alwahsh: well... This is kinda debatable :)
To me, it seems like one would always expect the result of an AJAX request that returns a 401 to redirect him to login page since it would be the same effect he'd get using normal HTML request.
Your opinion still makes sense though that one might want to handle it differently but I just can't think of a real use-case for that now. Let me apply your suggestion anyways :)

rohitpaulk: I'm okay if you want to leave this as it is - if and when a use case does pop up, we can change it :)

Alwahsh: Nope, I did change it... Just modifying the tests :)

app/controllers/relationships_controller.rb
```diff
@@ -4,7 +4,7 @@ class RelationshipsController < ApplicationController
 
   def identify_user
     @user = User.find_by(username: params[:id])
-    # Check for nil user and user trying to follow/unfollow herself
+    # Check for nil user and unsigned user trying to follow
     render js: "window.location = '/'" if @user.nil? || !user_signed_in?
```
rohitpaulk: We don't need this anymore.

rohitpaulk: Actually, we should render a 404 if the `@user` is not found.

PR #276, opened 2015-03-22
Follow layout changed and profiles linked

mecyborg:
Linked user profiles.
Changed layout of follower/following show.
Replaced current user show with "You".

### DIFF
--- a/app/views/users/_followers.html.haml
+++ b/app/views/users/_followers.html.haml
@@ -1,6 +1,10 @@
 %article
   -if @user.followers.empty?
     %p No one is following #{@user.username} yet.
-  - @user.followers.each do |follower|
-    %p= avatar(follower.email)
-    = follower.username
+  -@user.followers.each do |follower|
+    - if follower.username == current_user.username
+      = avatar(current_user.email)
+      = "You"
+    - else
+      = avatar(follower.email)
+      %a{href: "/#{follower.username}"} #{follower.username}
--- a/app/views/users/_followings.html.haml
+++ b/app/views/users/_followings.html.haml
@@ -1,6 +1,10 @@
 %article
   -if @followings.empty?
     %p #{@user.username} is not following anyone yet.
-  - @followings.each do |following|
-    %p= avatar(following.email)
-    = following.username
+  -@followings.each do |following|
+    - if following.username == current_user.username
+      = avatar(current_user.email)
+      = "You"
+    - else
+      = avatar(following.email)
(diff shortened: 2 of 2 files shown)


app/views/users/_followers.html.haml
```diff
@@ -1,6 +1,10 @@
 %article
   -if @user.followers.empty?
     %p No one is following #{@user.username} yet.
-  - @user.followers.each do |follower|
-    %p= avatar(follower.email)
-    = follower.username
\ No newline at end of file
+  -@user.followers.each do |follower|
```
Alwahsh: Can you put back the space you removed from that line? :)

app/views/users/_followers.html.haml
```diff
@@ -1,6 +1,10 @@
...
+    - if follower.username == current_user.username
```
Alwahsh: Better move that if condition one line down... You'd be printing the avatar of `follower.email` anyways

app/views/users/_followings.html.haml
```diff
@@ -1,6 +1,10 @@
 %article
   -if @followings.empty?
     %p #{@user.username} is not following anyone yet.
-  - @followings.each do |following|
-    %p= avatar(following.email)
-    = following.username
\ No newline at end of file
+  -@followings.each do |following|
```
Alwahsh: Give back the space please!

Alwahsh:
Having to do the same change in more than 1 place is a warning!
A good idea would be to use only one of those views with a local variable and assign that variable to its appropriate value in the calling file.

app/views/users/_followings.html.haml
mecyborg: Thanks for pointing that out. :)

mecyborg:
How can I miss that ?! Let me fix this. :D

app/views/users/_followers.html.haml
```diff
@@ -1,6 +1,10 @@
...
-  - @user.followers.each do |follower|
-    %p= avatar(follower.email)
-    = follower.username
\ No newline at end of file
+    
+  -@user.followers.each do |follower|
+    = avatar(follower.email)
+    - if follower.username == current_user.username   
+      = "You"
+    - else
```
sonalkr132: Hey, why have you used `if follower.username == current_user.username`? A user can never follow himself (if you found a case where he was then we would need to fix it)  
Also, I don't know if you noticed but you are failing some of them haml-lint rules. You might want to fix it.

PR #277, opened 2015-03-23
Fix JS tests on Travis

rohitpaulk:


.travis.yml
```diff
@@ -7,4 +7,4 @@ before_script:
   - RAILS_ENV=test bundle exec rake --trace db:migrate
 script:
   - bundle exec rubocop --config .rubocop.yml app/models/ app/controllers/ app/mailers/
-  - RAILS_ENV=test bundle exec rspec
+  - RAILS_ENV=test xvfb-run bundle exec rspec
```
Alwahsh: I remember adding this - Although not sure I put it in this exact place - but it didn't work :open_mouth:

Alwahsh: Ah found it.. I didn't actually put the xvfb-run in the right position. Great to see the tests working now :)

Issue #258, opened 2015-03-12
Add a code style linter to Travis


rohitpaulk:
This will involve a lot of work correcting the existing codebase, but it'll be worth it.

The most popular seems to be https://github.com/bbatsov/rubocop. We won't need all the Rails specific checks, just basic code style.

sonalkr132:
I know I need it more badly that anyone -_- 
Any useful links? also we should have wiki for coding style and best practices for noobs like me :stuck_out_tongue_closed_eyes:

rohitpaulk:
> Any useful links?

https://github.com/bbatsov/ruby-style-guide

sonalkr132:
yeah I saw that. Actually:

![screenshot from 2015-03-15 12 58 23](https://cloud.githubusercontent.com/assets/7680662/6655300/04f8ea38-cb1f-11e4-9b6c-0abe08f7a878.png)


That is rubocop with all its defaults.we should discuss about defining our own .rubocop.yml file? using [this](https://github.com/openSUSE/osem/blob/master/.rubocop.yml) offenses were reduced to 700

EmilyDirsh:
I'd prefer the rubocop defaults, since that uses the popular style guide that @rohitpaulk referred to above. It's more work to correct the code base, but it means less work defining and maintaining our own style guide, and will ultimately make it easier for new contributors.

sonalkr132:
I was thinking may be first we can write a simple .rubocop.yml, which counts only indentation and other textual offenses and not the coding style. For example by default it counts too long methods as offense, now re factoring them would be quite and involved approac and we have got a good number of long method definitions. Besides that it also counts no documentation as offense, another quite involved process. 
I am not saying we should throw away the defaults for good, I am just saying that we should start from something simple, get our code to pass that and work our way up from there.

EmilyDirsh:
@sonalkr132 I think it's fine to start small - just fix indentation, etc, but I'd like to maintain all the default checks still. For one thing, it's very likely that we'll all forget to go back and add the remaining checks in the future. For another thing, it'd be good to have all new contributions match the default, stricter criteria.

sonalkr132:
I would still suggest, turning some of the offenses off for now would be helpful for anyone who is fixing it. Don't worry that we won't go back to fix the rest of issues, you will have to take my word for it. Besides that we don't need everything of rubocop, for ex: it requires that every class definition has a comment block above it, I don't think anyone does that for rails apps.
[...]

PR #278, opened 2015-03-24
Rubocop spec

sonalkr132:
I do need a few suggestions. Does this look readable? Those 3 closing brackets one after look weird.
```Ruby
expect(Rugged::Repository.new(File.join(
      @project.data_path,
      'satellite',
      '.git'
    ))).to be_a(Rugged::Repository)
```

I had to change name of argument `password_confirmation` to `password_confirm`  here:
```Ruby
def sign_up_with(email, username, password, password_confirm = password)
  visit '/'
  click_link 'Sign Up?'
  # more_code
end
```
Both cases were violating line length limit. How could I have done better?


Alwahsh:
LGTM :+1:

rohitpaulk:
> Those 3 closing brackets one after look weird.

You could split the file_name out to a variable, and use that in the statement.

> I had to change name of argument password_confirmation to password_confirm here:

That's fine. 

Great work, thanks! I'm going to merge this.

sonalkr132:
How much priority should we give to this:
screenshot from 2015-03-24 21 36 06

rohitpaulk:
> How much priority should we give to this:

I think all those warnings are valid, and should be taken care of. This isn't very different from the cleanup we've been doing till now.

sonalkr132:
I want to get this cleaning code thing out of our way, I am gonna work on this too. I mentioned it [here](https://github.com/glittergallery/GlitterGallery/issues/258#issuecomment-84150580),  so just double checking :yum:

sonalkr132:
I strongly suggest that we change default line width to 100 for haml files. 
screenshot from 2015-03-25 00 59 04

Top one is not only has lesser readability but association between those two lines is also obscure.

rohitpaulk:
> I strongly suggest that we change default line width to 100 for haml files. 

Makes sense.

PR #282, opened 2015-03-29
Rubocop passed

### LINKED ISSUE #258: Add a code style linter to Travis

sonalkr132:
closes #258?

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,10 @@
 AllCops:
+  Exclude:
+    - 'db/schema.rb'
+    - 'config/**/*'
+    - 'lib/**/*'
+    - 'bin/**/*'
+    - 'Guardfile'
   StyleGuideCopsOnly: true
 
 Metrics/MethodLength:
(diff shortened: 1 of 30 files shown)


rohitpaulk:
Great work! I'm not going through the diff, gonna rely on the passing tests for now and get this merged in :)

sonalkr132:
What you should check is `.rubocop.yml` file:
```YML
AllCops:
  Exclude:
    - 'db/schema.rb'
    - 'config/**/*'
    - 'lib/**/*'
    - 'bin/**/*'
    - 'Guardfile'
    - 'vendor/bundle/**/*'
    - 'bundle/**/*'
```
We didn't generate any of those files, hence I excluded those. Is there any which should be included?

rohitpaulk:
Yes, I saw the `.rubocop.yml` fine, it looks great. 

> We didn't generate any of those files, hence I excluded those. Is there any which should be included?

Nope, that's fine. Even if we do manually edit a few of those, code style there isn't important.

PR #283, opened 2015-03-30
markup added to isssues and fix #191

### LINKED ISSUE #191: [Security issue] XSS in "Create Comment" box
Usually all the input are properly sanitized before being processed. 
But this doesn't seem to do so.
xss

For more details,
https://www.owasp.org/index.php/Cross-site_Scripting_%28XSS%29
https://www.owasp.org/index.php/XSS_%28Cross_Site_Scripting%29_Prevention_Cheat_Sheet

I hope you would understand the importance of this issue ;)

sonalkr132:
Do we really need to support html tags? I am have set redcarpet to filter it, but if we really need it then I guess I can add that support. [Here](https://github.com/vmg/redcarpet/issues/434) it explains how html support can be added.

UPDATE:
Why bother about glitterpost? I want to reach no security warning on hakiri and changes I just made fixed few of XSS warnings, hence glitterposts!

**#In_Progress**

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -45,4 +45,14 @@ module ApplicationHelper
     log_status = user_signed_in? ? 'logged_in' : 'not_logged_in'
     "#{controller.controller_name} #{controller.action_name} #{log_status}"
   end
+
+  # render markdown text on redcarpet
+  def markdown(text)
+    render_options = { hard_wrap: true, filter_html: true }
+    markdown_options = { autolink: true, no_intra_emphasis: true }
+    markdown = Redcarpet::Markdown.new(
+      Redcarpet::Render::HTML.new(render_options), markdown_options
+    )
+    markdown.render(text).html_safe
+  end
 end
--- a/app/assets/stylesheets/_markdown.scss
+++ b/app/assets/stylesheets/_markdown.scss
@@ -1,0 +1,75 @@
+@media screen and (min-width: 0em){
+  body{
+    .wrapper{
+      article{
+        section{
+          div{
+            div.option{
+              div.markdown{
+                h1, h2, h3, h4, h5, h6{
+                  background-color: #f9fafb;
+                  font-weight: bold;
+                  padding: 0.25em 0 0.25em 0;
+                  border-bottom: 1px solid #eee;
+                }
+                h1{
+                  font-size: 2.25em;
+                }
+                h2{
+                  font-size: 1.75em;
+                }
+                h3{
+                  font-size: 1.5em;
+                }
+                h4{
+                  font-size: 1.25em;
+                }
+                h5{
+                  font-size: 1em;
+                }
+                h6{
+                  font-size: 0.75em;
+                }
+
+                ul{
+                  padding-left: 1.25em;
+                }
+
+                ol{
+                  padding-left: 1.25em;
+                }
+
+                blockquote{
+                  padding: 0 15px;
+                  color: #777;
+                  border-left: 4px solid #ddd;
+                }
+
+                p{
+                  margin: 2% 0 2% 0;
+                }
+
+                pre{
+                  padding: 2%;
+                  overflow: auto;
+                }
+
+                code{
+                  font-family: Consolas, "Liberation Mono", Menlo, Courier, monospace;
+                }
+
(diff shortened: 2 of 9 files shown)


app/helpers/application_helper.rb
```diff
@@ -45,4 +45,14 @@ def set_body_class
     log_status = user_signed_in? ? 'logged_in' : 'not_logged_in'
     "#{controller.controller_name} #{controller.action_name} #{log_status}"
   end
+
+  # render markdown text on recarpet
```
rohitpaulk: small typo here, this should be 'redcarpet'

sonalkr132:
@rohitpaulk Thanks! Fixed it.

rohitpaulk:
@sonalkr132 - can you add tests that confirm that the markdown renderer indeed produces HTML tags? Also, how do the comments look with markdown? Do we have markdown specific styling?

sonalkr132:
following text:
```
I am **bold**
I am *italic*
`I am code`
# I am head
[I am link](https://google.com)
> I am quote
```
Gets translated to:
```
<p>
  I am <strong>bold</strong><br>     
 I am <em>itallic</em><br> 
 <code>I am code</code><br>
 I am <h1>head</h1>
 <a href="https://google.com">I am link</a>
<blockquote> I am quote </blockquote>
</p>
```

So pretty much like github. Redcarpet uses [Daring Fireball](http://daringfireball.net/projects/markdown/syntax), so other markdown listed there should work too (except html markdown, I am filtering those). 
However, `* list` doesn't seem to work.

rohitpaulk:
Great, we'll need tests to confirm that.

sonalkr132:
We need to find an answer to [this](http://stackoverflow.com/questions/29347964/daring-fireball-list-markdown-not-working) first :stuck_out_tongue_closed_eyes:

sonalkr132:
I hope you are not concerned about indentation. I am not violating guideline, it was either that or lot of `\n`.

rohitpaulk:
> We need to find an answer to this first

http://stackoverflow.com/a/29348286/3189203

rohitpaulk:
The implementation looks good to me, I'm still concerned about the styling though - do we have markdown specific styling? We don't want a `h1` in the markdown looking like it's as important as a `h1` that we generate.

sonalkr132:
> We don't want a h1 in the markdown looking like it's as important as a h1 that we generate.

By **important** do you mean important in terms of SEO? As in we won't want search engines to get confused with those `h1` tags and it would be a bad html markdown? 
I don't think the gem provides a way to configure the tags it uses for styling. Of course we can do it anyway.. parsing the input with nokogiri can be one solution. But do we really need it?

once again I would like to way things are done github. Inspect the thingy below.
# HEAD 

You will see that it is using `h1` tag to style.

rohitpaulk:
I'm not talking about seo, I'm talking about visual importance. 

Every site that embeds user-provided markdown has markdown specific styling. For example, the h1 in github is styled by `.markdown-body h1`.

sonalkr132:
oh I get it now. yup! I just checked, styles we have for `h1` is interfering with markdown added `h1`. Would adding the class to `h1` of pages (not `h1` markdown) and using that as selector be good enough?

sonalkr132:
Okay.. it is much more complicated than just adding an extra selector. we can't use a lot which we have right now. for ex:
```Ruby
%h2.issue
   = markdown @issue.description
```
or 
```Ruby
%p
   = markdown issue.description 
```
`h1` tag can't be contained within `p` or `h1` tag. Besides thats:
screenshot from 2015-04-15 20 58 22
Looks like we will need to define style specific to markdown with something like you pointed out `.markdown-body`. I will add commits to this, just not sure when!

sonalkr132:
@rohitpaulk I have added style. it looks something like:
teamrocket

#254 needs to be reviewed for me to work on its style as right now comment use textfield which obviously can't support markdown. I have changed it to textarea there.

rohitpaulk:
This looks great, @sonalkr132! 

#254 is quite a bit to review, I'll have a look at it this weekend (unless @sarupbanskota gets to it first :wink:)

sonalkr132:
**update:**
Markdown specific style added to comments as well.

PR #284, opened 2015-03-31
Solve XSS on hakiri

sonalkr132:
We need to do something about routes files -_- Any pointers?

### DIFF
--- a/app/views/issues/show.html.haml
+++ b/app/views/issues/show.html.haml
@@ -42,7 +42,7 @@
           %nav
             %ul
               %li.active
-                = link_to('Close', File.join(@issue.show_url, 'close'), method: 'post')
+                = link_to('Close', "#{@issue.show_url + '/close'}", method: 'post')
               %li
                 Merge
 
--- a/app/views/projects/file_history.html.haml
+++ b/app/views/projects/file_history.html.haml
@@ -1,28 +1,32 @@
-- if private_check
-  - content_for :title do
-    = @project.name.titleize
-    File history
+%article.project
+  %header
+    %h1
+      = avatar @user.email
+      = @user.username
+    %h2
+      = precede '@' do
+        = @user.username
 
-  - content_for :pagetitle do
-    = link_to @project.name.titleize, @project.urlbase
-    \/
-    = link_to params[:image_name], File.join(@project.urlbase, 'master', params[:image_name])
-    \/
-    File history
+  %section
+    %div
+      - if private_check
+        - content_for :title do
+          = @project.name.titleize
+          File history
 
-    = render 'project_toolbar'
-    = render 'shared/messages'
+        = render 'project_toolbar'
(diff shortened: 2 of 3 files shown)


Alwahsh:
Nice description you have there :)
Don't worry about the routes file... I'm working on the rest of it... I didn't yet get to the issues and pull requests parts.
(done some part of the issues but didn't upload it yet)
If you see something wrong with other parts. Let me know but please be as descriptive as possible if you want me to understand you.

sonalkr132:
Sorry! 
Do you have any idea why we have two routes for everything? for ex:
```Ruby
  post '/:user_id/:id/issues/new' => 'issues#create'
  post '/:user_id/:id/issue/:sub_id/close' => 'issues#close'
```
and
```Ruby
  post '/:user_id/:id/:xid/issues/new' => 'issues#create'
  post '/:user_id/:id/:xid/issue/:id/close' => 'issues#close'
```
What does that small `:xid` do?

Alwahsh:
That xid is for private projects... As I said, I didn't yet refactor the issues part so no need to look at their routes now.

sonalkr132:
In case you are looking for things which are wrong with routes, I think we should replace all the links which have things like:
```Ruby
%a{ href: "#{@project.urlbase + @project_path.to_s}" }
    %div{ data: 'current' }
      %p
        Current
```
to the rails way ie `link_to Current, user_project_path(@user, @project)`. Don't you think?

sarupbanskota:
@Aditya - sure, when you see something fit, just change it :-)

sonalkr132:
Changing things to rails way is lot of work and would require dedicated effort rather than changes on go. Things going on in the controller depends a lot on what we send in `params`. So besides changing the routes we would also need changes in controller. For ex: lines like `@issue = Issue.find_from_project(@project, params[:sub_id])` won't be required if use routes rails way. methods like `return context Doesn't performing lot of queries bases on what user sends as params leaves us vulnerable? SQL injections?

sonalkr132:
Our css files would also change, one other reason which won't allow us to make changes on go. For ex:
```
%a{ href: "#{@project.urlbase + @project_path.to_s}" }
    %div{ data: 'current' }
      %p
        Current
```
css is defined for `a div p` but if we use `link_to` then we would need to use different selector. Since styles are shared by many, we can't make changes to one of them or for one them but we will need to everything at one.

I guess what I am trying to ask is.. will changing to `link_to` and `*_path` be worth it?

sarupbanskota:
yup, essentially, if we're changing something, then we want to also fix
everything that breaks because of that change. if the change isn't
absolutely essential + it breaks more things than we have time to fix, then
we skip.

ofc, doing thing the rails way doesn't hurt, so.. based on the time you
have + what you're working on, your call :)



Sarup

sonalkr132:
No longer relevant/up-to date with the master

PR #288, opened 2015-04-03
remove user email from path

sonalkr132:
If someone "opens image in new tab" (there are other ways too) then he would see something like :`http://localhost:3000/data/repos/aditya.prakash132@gmail.com/fancy_project/satellite/10407084_993180024044843_1803483709906686358_n.png`. I don't think we should keep user email public like that without permission of user.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -150,7 +150,7 @@ class Project < ActiveRecord::Base
   def set_path
     user = User.find user_id
     self.data_path = File.join Glitter::Application.config.repo_dir,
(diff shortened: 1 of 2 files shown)


sarupbanskota:
yup, good catch.

i'm wondering if this change won't break anything?

Alwahsh:
@sarupbanskota I've made thjs change in my demo version(needed it for something else) and as far as I can see, it didn't break anything :)
and they're passing tests so that's more confirmation :)

sarupbanskota:
alright, let's merge then.

Issue #220, opened 2015-03-06
Validate file uploads


EmilyDirsh:
Currently file uploads are not validated. We should be restricting the type and size of files permitted to be uploaded.

This is a potential security issue, and can certainly cause a lot of usability issues if we try to treat something as an image that isn't.

Relevant: http://stackoverflow.com/questions/51572/determine-file-type-in-ruby

sbagroy986:
On it!

sbagroy986:
Found a gem for it. What file formats should I add as supported @EmilyDirsh ? JPEG and PNG?

Alwahsh:
Why not make it configurable ?

sbagroy986:
I don't understand what you mean. Configurable how?

Alwahsh:
For example, an Array with accepted formats put in the configuration file in initializers. Allow everything in that array and reject the rest

sbagroy986:
That is what I am doing :)

EmilyDirsh:
@sbagroy986, @Alwahsh is right - the whitelist should be configurable in the Rails config files. Check the documentation for the gem you're using - there should be a documented configuration parameter for the gem.

By default, we should allow jpg, png, gif (why not?), and svg

sbagroy986:
The gem actually just provides us the format of the file. So basically I'm gonna create an array of acceptable formats. The list is, thus, configurable.

EmilyDirsh:
@sbagroy986 which gem are you using?

sbagroy986:
Ruby FileMagic - https://github.com/blackwinter/ruby-filemagic
Its basically libmagic for ruby.

sarupbanskota:
What's the status on this? If it's yet to be worked on, @sonalkr132 would like to take it up.

PR #289, opened 2015-04-03
fix #287: update image working

### LINKED ISSUE #287: File update method in projects controller
Here is the method: [projects_c/line272](https://github.com/glittergallery/GlitterGallery/blob/master/app/controllers/projects_controller.rb#L272)
and here the satellite method it is calling [application_c/line55](https://github.com/glittergallery/GlitterGallery/blob/master/app/controllers/application_controller.rb#L55)
This is driving me insane -_- lets begin!
`satellite_commit` has 3 arguments in definition but while we are calling it in `file_update` we are passing 4 arguments. okay! lets assume that only three were meant to be passed and lets remove `params[:image_name]`. So it would look something like:
```
satellite_commit @project.satelliterepo,  imagefile.read, message
```
`imagefile` is an object of `ActionDispatch::Http::UploadedFile`. `read` on an object of `ActionDispatch::Http::UploadedFile`? okay may be we meant pass just `imagefile`. moving to `satellite_commit`:
```
files.each do |f|
```
`each` on object of `ActionDispatch::Http::UploadedFile`? May be we meant to pass `tempfile` in `satellite_commit`? Now you pass `imagefile.tempfile` instead? but hey! `read` works on `temfile` so may be you meant to pass `imagefile.tempfile.read`? I wish I had a clue. Lets assume you have passed whatever you meant to.
lets move to `satellite_commit` again. 
```
files.each do |f|
      tmp = f.tempfile
```
Next line says ` tmp = f.tempfile`. We just agreed that may be we meant to pass `tempile`, sure thing we can't do `f.tempfile` on `tempfile`. From little I have understood so far, looks like it is support for upload of multiple images. Although a user is not able to select multiple files to upload in UI(use ctrl btn). Both `file_update` and `satellite_commit` has lines:
```Ruby
file = File.join @project.satellitedir, f.original_filename
FileUtils.cp tmp.path, file
```
Why are we saving twice? Are we using `satellite_commit`any where else in our app? Why is it defined all the way in `application_controller`? This would the time when I lost it. :sob: 
Please help!!!! What sorcery is going on here?

sonalkr132:
closes #287 (hopefully)

Of course we need to add tests. Before that:
* When user click on `update` he is updating a single image, hence I am not allowing him to select multiple images. However, I am using an array of `ActionDispatch::Request` so that we can use same function `satellite_commit`. Am I right?
* When user update an `image1.jpg` with `image2.jpg`. We keep both copies of `image1` and `image2` in tree. Shouldn't we remove `image1.jpg` from tree?
* Can I take up #220? We already have #236 but there seems to be no activity on it. #220 needs to be fixed for me to write tests.

**#In_Progress** **#Need_Feedback**

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -270,15 +270,11 @@ class ProjectsController < ApplicationController
   end
 
   def file_update
-    tmp = params[:file].tempfile
-    file = File.join @project.satellitedir, params[:image_name]
-    FileUtils.cp tmp.path, file
     if params[:file]
       imagefile = params[:file]
       message = params[:message]
       commit_id = satellite_commit @project.satelliterepo,
-                                   params[:image_name],
-                                   imagefile.read,
+                                   imagefile,
                                    message
       generate_thumbnail @project, params[:image_name], commit_id
       @project.pushtobare
--- a/app/views/projects/update.html.haml
+++ b/app/views/projects/update.html.haml
@@ -10,8 +10,8 @@
     %section
       %div
         = form_tag file_update_user_project_path(@project.user, @project), multipart: true do
-          = file_field_tag 'file'
-          = text_field_tag 'message', 'commit message'
+          = file_field_tag 'file[]'
+          = text_field_tag 'message', nil, placeholder: 'commit message'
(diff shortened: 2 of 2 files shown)


sonalkr132:
Fixed in #295

PR #291, opened 2015-04-05
fix style of forgot password

sonalkr132:
Sorry, for redundant PR  there was some issue with github:
> Sorry for the trouble. We deployed a change that seems to have created some PR buginess.
> We're working on a fix, and I'll let you know when we have an update.

Anyway, I had to add class because I wanted to style `sign in` and `sign up` links at bottom. `_form.scss` didn't have style defined for them and it was appearing as normal links (blue underline ones). I thought it would be better if forgot password pages' style are standalone and defined in a scss file of its own.

I have removed the class and instead added the style to `_form.scss` file.


app/assets/stylesheets/_form.scss
```diff
@@ -1,99 +1,115 @@
...
+            padding: 0;
+            text-align: left;
+            p{
+              a{
+                color: $map_blue;
+              }
+              a:hover{
+                font-weight: 700;
+              }
+            }
```
sonalkr132: Style in `section.option` would be the ones I defined. It is kind like ones defined `_hero.scss` but not exactly. It has different width, padding for `div` and text-alignment and padding for `section.option` than that of `hero`, hence inheriting didn't seem like  great idea.

Issue #285, opened 2015-03-31
Issue create redirects to dashboard of projects


sonalkr132:
Create method of Issue has:
```Ruby
redirect_to(dashboard_path)
```
Shouldn't it redirect to issue show page?

Alwahsh:
yeah... I actually made it work like that in a PR that I didn't yet upload :)

sonalkr132:
oh can you reference PR here? I will close it.

Alwahsh:
I don't think you should close it... They might have some reason for making it that way so I suggest you leave this open till they confirm.

mecyborg:
I was also wondering why its like that but it can be intentional, so please wait till they confirm.

sarupbanskota:
It should redirect to issue page, correct.

Sarup

Alwahsh:
This can be closed now.

sonalkr132:
fixed: #292

PR #292, opened 2015-04-05
Refactored issues, project images browsing, added branching support and other minor improvements

Alwahsh:
I've refactored the issues part and done some modifications, following is what I recall:
* Modified the routes to be more rails way.
* Removed the merge button from issues(I found it useless. I can't imagine a way where one can merge issues anyway)
* Report new issue button would not appear if there's an issue already, I fixed that(It was commented out so all I did was to uncomment it.)
* One can now reopen an issue.
* Made adding a new issue redirect you to the created issue. See #285.
* Added tests for that part. Now we're at 90% test coverage.
* Comments would be posted but no javascript response was rendered. I fixed that(Things needed were already there but just not correctly connected.)
* Comments appeared in a cool way only in the logs page... I've now made them like that on issues and others.
* Posting a new comment on an image file would actually post a comment to the project (current page). Fixed that.
* Added rubocop to guard. I made this in the second commit since I felt it's undrelated to improvements.

I can't remember a lot but as I browsed the files, I fixed whatever I felt wrong :)

Note that in this I have REMOVED the merge button from issues as noted above. If you want me to put it back please let me know... Also, there was a commented "Star" option for projects. I REMOVED that part completely. I think the "Follow" for projects does what "Star" was made for.

Also, I recognize there's #267 which improves the issues... I didn't want to duplicate the work there but I modified the issues model a little bit so that my tests would pass when #267 gets merged.

I didn't edit any authorization issue till now. Waiting for #244 to get merged to be able to have well defined abilities as any work done now would be overwritten later.

PR #293, opened 2015-04-07
Tagging

sonalkr132:
I added [acts-as-taggable-on](https://github.com/mbleigh/acts-as-taggable-on) to issues. Surprisingly enough I passed rspec.  I have added sorting to tags as well.

* Do projects need to be taggable too? 
* How far should we take the styling of tag? We can get by changing delimiter of tags from comma to something else and autocomplete of jquery or do we need to something like [chosen-rails](https://github.com/tsechingho/chosen-rails) or [select2-rails](https://github.com/argerim/select2-rails)?

#In_progress #Needs_feedback

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,6 @@
 AllCops:
   Exclude:
-    - 'db/schema.rb'
+    - 'db/**/*'
     - 'config/**/*'
     - 'lib/**/*'
     - 'bin/**/*'
--- a/Gemfile
+++ b/Gemfile
@@ -50,6 +50,8 @@ gem 'therubyracer', platform: :ruby
 
 gem 'uglifier'
 
+gem 'acts-as-taggable-on', '~> 3.4'
+
 
 group :development, :test do
   # https://github.com/rspec/rspec-rails/issues/1273
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -33,6 +33,8 @@ GEM
       minitest (~> 5.1)
       thread_safe (~> 0.1)
       tzinfo (~> 1.1)
+    acts-as-taggable-on (3.5.0)
+      activerecord (>= 3.2, < 5)
     addressable (2.3.7)
     ancestry (2.1.0)
       activerecord (>= 3.0.0)
@@ -317,6 +319,7 @@ PLATFORMS
 DEPENDENCIES
   actionpack-action_caching
   activerecord-session_store
+  acts-as-taggable-on (~> 3.4)
   ancestry (~> 2.1.0)
   better_errors
   binding_of_caller
--- a/app/controllers/issues_controller.rb
+++ b/app/controllers/issues_controller.rb
@@ -7,11 +7,11 @@ class IssuesController < ApplicationController
   end
 
   def index
-    @activetab = 0
-    @issuestoshow = @project.issues.where(status: 0)
-    return unless params[:state] == 'closed'
-    @activetab = 1
-    @issuestoshow = @project.issues.where(status: 1)
+    @issuestoshow = find_issue(params[:state])
+    if params[:tag]
+      @activetab = nil
+      @issuestoshow = @project.issues.tagged_with(params[:tag])
+    end
   end
 
   def new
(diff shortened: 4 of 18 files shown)


sarupbanskota:
@sonalkr132 let projects not be taggable for now.

about the styling, let's stick with comma + autocomplete. we can revisit when we want to improvise :)

Alwahsh:
>  Surprisingly enough I passed rspec

Why is that surprising? The issues part is not covered by any tests yet since #292 is not yet merged and you didn't add any tests so there's no way you can fail rspec :)

I see you've worked on some of the stuff I worked on too... So this would close #285 

Also, a question... if you've defined a list of accepted tags, wouldn't it make more sense to make the user select one of them instead of leaving him a textbox where he writes any tag he wants... I tried this locally and I couldn't add any issue and never got a clue on why's that except when I saw the line where you force the tag to be one of (bug feature improvement inspiration).. This would be a really bad experience for users since they can't magically guess which tags you hardcoded in the code.

Alwahsh:
@sarupbanskota I'm not sure if my above comment would be an enough reason, but I think this should be reverted at least till the issue I mentioned above is fixed and till we have some tests?

rohitpaulk:
I'm reverting this - We really need tests for all this functionality.

sarupbanskota:
Yep I agree - thanks for reverting it Paul.

On Thursday, April 9, 2015, Rohit Paul Kuruvilla <notifications@github.com>
wrote:

> I'm reverting this - We really need tests for all this functionality.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/glittergallery/GlitterGallery/pull/293#issuecomment-91148252>
> .
>


-- 
Sarup

sonalkr132:
hey! why would you merge this in first place. I need feedback, I had work more on this and thats why:
> #In_progress #Needs_feedback

I haven't added the style or autocomplete. I hadn't written any tests. As @Alwahsh pointed out, I wasn't sure if user can make any tag they want or we should have limits and I added a limited version because in comments it said `TODO: find 5 most popular tags among designers`.
> Surprisingly enough I passed rspec. 
I meant even after such a big change I passed rspec which is surprising and something is obviously wrong and we need to fix it. ie write tests.

Alwahsh:
@sonalkr132 I think it's better to indicate in the title when your PR is in progress. 
Also... The number of PRs that you have in progress is increasing... Why not focus on one thing till it's complete? you're just making it harder -for yourself and for me- by giving more chances of conflicts like that... If you had opened this PR before I opened #292 , you'd have blamed me for working on issues.

>I meant even after such a big change I passed rspec which is surprising and something is obviously wrong and we need to fix it. ie write tests.

yup.. We have no tests for the issues part as I mentioned and fixed in #292.

sonalkr132:
@Alwahsh 
> it's better to indicate in the title when your PR is in progress. 

Yeah, thats a nice idea. 
> Why not focus on one thing till it's complete? 

I leave them in progress cause I am waiting feedback on them. Out of 7, 2 are in progress so not that many I would say.
> We have no tests for the issues part as I mentioned and fixed in #292.

My finals are close -_- I can't work much. May you can pull this my branch on issues tagging? I will add auto complete and some style to tags. Or may if I have time, I will pull from yours. Don't use rebase or force update if I have pulled.

"Should the list be decided by one single admin or users should be able to add whatever tag they want on tags?" is still unanswered.

Alwahsh:
> May you can pull this my branch on issues tagging? I will add auto complete and some style to tags. Or may if I have time, I will pull from yours. Don't use rebase or force update if I have pulled.

I didn't really understand this paragraph fully but it would be easier for me to just add the gem myself and write the tests as I introduce the feature. Issues tagging is something that I planned to work on but not at this point. So, I prefer to put my effort in finishing the partly implemented but not yet fully working features before I introduce others for now.

> My finals are close -_- I can't work much. 

Yeah... I can feel you... While I still have some time before my finals but I have things to do in my college too (projects and practical exams :( ). It'd have been cool if they gave bonus for open source contribution :)

> "Should the list be decided by one single admin or users should be able to add whatever tag they want on tags?" is still unanswered.

Let me tell you my opinion... I believe whatever tags is better but one still gets a list of previously created tags as suggestions as he's writing. If he finds a suitable tag, he selects it. Otherwise, creates his own. Not sure how others would think of that though.

sonalkr132:
> I didn't really understand this paragraph fully

Sorry, I meant that either of us can pull from branch of other and they can just sent one PR. About adding tags.. I only started it because I didn't find anything else to work on. I think I I will work on this after your PR is merged.

> Otherwise, creates his own. 

Wouldn't it require that someone moderates tags? List of tags can be as extensive as it has to be. We can even write a wiki page about tagging. I think thats how it works on github?

PR #297, opened 2015-05-10
Tagging

sonalkr132:
Only validation of tags is their presence that would mean that user can add any tag they want. I am not sure how should I go from here. Who should control the types of tag here? Unlike github, user who creates an issue can also tag his issue, which may lead to some problems.

Also, let me know if I need to write more tests. 

**#Needs_Feedback**


sonalkr132:
I have no idea why travis build hasn't started, hopefully it will start soon.

I added autocomplete using jquery-ui. It also shows the complete list of tags when user focuses the tag list's text area:
screenshot from 2015-05-12 01 54 35
I hope style is good enough. Owner of the project can make new tags and general users can only use one of them. Few basic tags are added with seed, which will be present even if owner has not added any tag.

PR #298, opened 2015-05-16
Exploration page

sonalkr132:
- [x] give a project of rating 1-5 on both user-projects show page and project-show page
- [x] user-projects page will show average rating of project and project-show page will have rating of current user on that project
- [x] Write tests
- [x] make inspiration page into an exploration page with option to sort on basis of average ratings, number of followers, number of forks, most active in past 10 weeks (bases on no of commits and issue closed)
- [x] write test for sorting (needs help)
- [x] better image layout of project page (pending review)
- [x] write test for responsive images rendering on inspire page

**#Needs_Feedback ~~#In_Progress~~**

PR #300, opened 2015-05-25
authorization

sonalkr132:
Besides update image there were many more actions which were unprotected. I hope I have covered them all in right way. Let me know if it needs changes.

- [x] authorization to projects controller
- [x] tests for projects controller auth
- [x] update of view for projects
- [x] authorization to comments and issues controller
- [x] tests for comments and issues auth
- [x] update of view
- [x] add auth to user
- [x] write pending test for guest after #298 gets merged
- [x] make inspire as root

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,6 +54,8 @@ gem 'acts-as-taggable-on', '~> 3.4'
 
 gem 'ratyrate', github: 'sonalkr132/ratyrate', branch: 'refactoring'
 
+gem "cancancan", '~> 1.10'
+
 
 group :development, :test do
   # https://github.com/rspec/rspec-rails/issues/1273
--- a/Gemfile
+++ b/Gemfile
@@ -54,7 +54,7 @@ gem 'acts-as-taggable-on', '~> 3.4'
 
 gem 'ratyrate', github: 'sonalkr132/ratyrate', branch: 'refactoring'
 
-gem "cancancan", '~> 1.10'
+gem 'cancancan', '~> 1.10'
 
 
 group :development, :test do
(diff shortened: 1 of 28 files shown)


sonalkr132:
demo: http://demo-glitterfollowers.rhcloud.com/

Any chance that speed will be better if we use paid service? Cause right now it sucks. Something it will just keep loading you have to hit refresh.

sonalkr132:
#234, #232, #236 and #244 can be closed.

sonalkr132:
Check out heroku deployment: https://calm-sea-9315.herokuapp.com
It is at least 10x times. Except you need to look at it before the app sleeps or else heroku will delete the public data folder.

sarupbanskota:
I noticed one bug so far - it's not related to auth though. If you look at the screenie, unstarred projects seem to show up before 4 or 5 star projects when sorting by stars. 

Rest lgtm for now, so merging.

screen shot 2015-05-31 at 7 43 12 pm

sonalkr132:
Actually that is something related to db.  By default, null values sort as if larger than any non-null value; that is, NULLS FIRST is the default for DESC order, and NULLS LAST otherwise if we are using postgres (something I have use while deploying on heroku) but not when we use mysql or sqlite.

It is an easy fix, we can add NULL LAST to the sql query and null values will come as last in all the cases. There is one more problem with using pg. I have to change polycomment_id from string to integer every time, cause pg is strong type check db. 
I was keeping this discussion for meeting cause I am not sure we should worry about supporting all kind of db out there. However these two changes make sense and probably we should do them anyway.

Issue #161, opened 2014-09-11
Local Repo Support


EmilyDirsh:
As a designer using Glitter Gallery,
I want to work on my projects in a local git repo
   and push changes to Glitter Gallery using the git client of my choice,
So that I can spend less time uploading files through the web interface.

As a designer using Glitter Gallery,
I want to pull any files I create or edit in the Glitter Gallery web console into my local project repo
So that I can easily work on the files locally without having to manually download and reupload them.

sonalkr132:
Can be closed. All the mentioned features are now supported.

sarupbanskota:
Thanks <3

PR #302, opened 2015-06-10
grack and auth

sonalkr132:
- [x] integrate gitlab-grack
- [x] override auth of grack
- [x] allow cloning of only public projects
- [x] auth for cloning of private projects
- [x] auth for push to projects
- [x] change path of bare repo and sync it with non-bare one
- [x] generate commit and inspire images
- [x] test for auth before clone, push, pull and fetch
- [x] allow owner of project to add collaborators
- [x] test project access


sonalkr132:
Here is deployment: http://ruby-glitterfollowers.rhcloud.com

Please take out some time to test it and break it. I have covered all the cases I could think of, however I am sure there are cases left where it will break.
Points to look out for:
* User should be able to clone public projects. For ex: `git clone http://ruby-glitterfollowers.rhcloud.com/addie/cards.git` should work.However it will ask for credentials if you clone private ones.
* It will always ask you for credentials if you are going to push. Use `username` and not `email`
* If you have made changes form the web UI, make sure you pull those changes first or else do a force update
* Inspire page images will only change if you make changes to master branch (for simplicity)
* Images which are used for thumbnail and inspire image generation won't necessarily be the last image you update, it will use the last image which rugged:diff return. There is no notion of last updated file/image in git between two commits.

Possible changes/refactoring:
* **Right now only project owner has option to clone private project and make push** There is no option to add collaborators.
* support both username and email for login
* private projects show on inspire page
* refactor the way sync is done between non-bare and bare repo. I have overloaded grack::server's service_rpc method to achieve it. Intuitively it feels wrong, most likely because it makes the code brittle.
Help:
I haven't looked much into sparkleshare, I still have 10 days before my current iteration ends, so it should be enough time. First look tells me that it is job of sparkleshare that they support our website (like the way the have added support for github, gitlab, piano, bitbucket etc). Let me know what you guys already know about integrating sparkleshare.

sonalkr132:
#296, #301 and #302 can be merged and #161 can be closed. woot woot!

sonalkr132:
I have updated deployment http://ruby-glitterfollowers.rhcloud.com/

You can test following users for private project: addie/Batman
username: collaborator
password: secretpassword
should be able to clone/fetch/pull and push to http://ruby-glitterfollowers.rhcloud.com/addie/Batman.git

username: reporter
password: secretpassword
should be able to clone/fetch/pull but not push.

For further details of ability please read [spec/models/ability_spec.rb] (https://github.com/sonalkr132/GlitterGallery/blob/grack/spec/models/ability_spec.rb)

NOTE: You can see private projects on index page. <- will fix in a new PR

PR #305, opened 2015-08-18
Improvements of project show page

sonalkr132:
- [x] lightbox gallery

~~- [ ] support markdown file~~
- [x] Change database to postgres. More info: #307 
- [x] search project and user name (context: website)
- [x] search files and issues (context: project/repo)
- [x] Show number of issues, commits, branches and file types
- [x] Test of all the above
- [x] Responsive images for project show page


sonalkr132:
Fixed #276

sonalkr132:
Hey, check this:
screenshot from 2015-09-09 01 04 24
I was thinking may be I would show files and issues results (searched within project) on two separate tabs. But the position of tabs looks kinda awkward, they are right below toolbar which is just another set of tabs.
Suggestions?

sonalkr132:
I ended up using jquery-ui for tab switching. It is not an added load, we already are using it for slider. It looks something like this now:
screenshot from 2015-09-09 01 42 10

sonalkr132:
hey guys this is complete. I have updated the [deployment](http://glittergallery-dev.fedorainfracloud.org/) as well.

sonalkr132:
Sorry but project show page was awful. It got on my nerve -_- It took forever to load. I have added responsive images for it as well. As Emily has suggested I am generating them on fly, ie only when they are requested. It won't lead to bloat as I am deleting all the previous images before generating new images. Size of images used is same as that of images on inspire page.

PR #310, opened 2015-09-19
Repo file path improvements

sonalkr132:
commit thumbnails are broken on fork (check [here](http://glittergallery-dev.fedorainfracloud.org/sachinkamath/fancy_project/commits)) cause the folder was not copied from parent.

Yet to implement:
Move repos from public dir to outside rails app. It leaves private repos exposed to outside world as everything in public dir can be rendered directly. ie routing doesn't involve rails app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,5 +25,6 @@ tmp/
 [._]s[a-w][a-z]
 *.un~
 
-# Ignore database config file
+# Ignore env specific config files
 /config/database.yml
+/config/initializers/0_glitterconfig.rb
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ rvm:
   - 2.2.0
 before_script:
   - cp config/database.yml.example config/database.yml
+  - cp config/initializers/0_glitterconfig.rb.example config/initializers/0_glitterconfig.rb
   - psql -c 'create database travis_ci_test;' -U postgres
   - RAILS_ENV=test bundle exec rake --trace db:migrate
 script:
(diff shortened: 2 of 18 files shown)


sonalkr132:
State: working in development env

Wouldn't a yml file instead of initializer ruby file ([0_glitterconfig.rb.example](https://github.com/sonalkr132/GlitterGallery/blob/fork/config/initializers/0_glitterconfig.rb.example)) be a better place to store app settings?

sarupbanskota:
@sonalkr132 sorry about the massive(!!) delay in replying.

Any links I can tinker with to try this out?

sonalkr132:
There isn't much to see in deployment as from user's perspective this PR changes nothing.
Try visiting this link: http://glittergallery-dev.fedorainfracloud.org/home/git/repos/addie/private_ryan/show_images/desktop/v2-cute-cat-picture.jpg
You will be forbidden from seeing that cute picture of kitten unless you are logged in with "aditya.prakash132@gmail.com" and "secretpassword". Notice the change in url after domain name. Now it is: /home/git/repos (previously it was /data/repos/ which existed in app itself).

sonalkr132:
This can be merged now. I have updated the deployment. Purpose of this PR is that no one should be able to see images or content of git repo by bypassing the rails routing (which previously was possible cause everything existed in public dir of app). Check out this link (`secret` is a private repo):
http://glittergallery-dev.fedorainfracloud.org/home/git/repos/addie/secret/show_images/desktop/1xC2BYX.png
You will get a forbidden access error unless you have read access of the repo.

PR #311, opened 2015-10-02
Add annotation to images using annotorious

sonalkr132:
- [x] add non persistent annotation to blob view
- [x] fix style overrides on annotation elements
- [x] merge @Alwahsh/more-improvements to fix blob comments
- [x] make ajax call to store annotations in db
- [x] add a button to turn off annotations markups
- [x] allow user to see raw image files
- [x] tests


sarupbanskota:
When you have a sizeable demo running, be sure to post it here and I'll get your branch merged.

sonalkr132:
You can check annotation [here](http://glittergallery-dev.fedorainfracloud.org/addie/fancy_project/blob/master/ss-150129-twip-06_9c81240dd2e6928eca16cc389ec7968d.nbcnews-fp-320-240.jpg#). Sign in to make new annotation.  Click on the *annotation link* to see annotations. They are hidden by default. After that you just need to click and drag on image to make annotations. Let me know if something doesn't work.


Also, I am putting #310 **on hold**. I want #311 merged first.

sonalkr132:
About failing hakiri check. They have always been failing. I haven't introduced any new security issue. Hakiri checks on PR seems to be a new feature. Most of the critical/severe warnings can be handled just by update of rack and jquery-rails versions we are using. I don't want to make all those changes here.

sarupbanskota:
This LGTM - I verified on the link you provided, so I've merged it. Thanks @sonalkr132!

PR #332, opened 2016-01-15
Fixes on some of the features

sonalkr132:
- [x] File with same name can't be uploaded
- [x]  Use of deleted project name throws error


sonalkr132:
I have dropped support of reuse of deleted project names here. Allowing it would lead to unnecessary ambiguity (and of course indexing error). We can allow recovery of deleted projects, that may be a story for a different PR.

PR #333, opened 2016-01-26
Introduce polling for live update of messages and notifications

sonalkr132:
- [x] update to rails 4.2.5.1
- [x] live messages
- [x] live notification
- [x] tests


sonalkr132:
Guide followed for update to rails 4.2.5.1: http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-1-to-rails-4-2

`web-console` and `responders` was not added cause we don't really need it. We might consider introducing `spring`, it has been included by default since rails 4.1, however rails 5 seems to have issues it for now.

One of the crucial updates is update of devise to 3.5.5 due to security vulnerable. You can read about it [here](http://blog.plataformatec.com.br/2016/01/improve-remember-me-cookie-expiration-in-devise/).

sonalkr132:
Phew! Finally travis is green. Drop in test coverage is due to update of `simplecov`. And they are right, some of our files have 0% which weren't included in test coverage previously, they are included now.
However, I won't be working on test coverage in this PR.

app/assets/javascripts/comments.js.coffee
```diff
@@ -0,0 +1,17 @@
+@CommentPoller =
+	poll: ->
+		setTimeout @request, 30000
```
Alwahsh: How about putting that number in some variable for configuration? I don't think it should be hard coded.

sonalkr132: Thanks @Alwahsh . I have made the required change.

